PR #4, opened 2025-04-01
Fix panic when using multi-byte character after shortcut link

SeijiIizuka:


rvben:
Hi @SeijiIizuka ,

Thanks for the PR. This resolves another edge case I didn't know about. :)


Kind regards,

Ruben

PR #130, opened 2025-11-01
feat(cli): Add server config command for LSP configuration inspection

elementalvoid:
😬 Hey there, sorry, this is a bit spam'ish. I am not a Rust dev and so have been using Copilot to help understand this application's structure and work on:
- a fix for what looks like a bug to me (lsp `server` command not loading configs the same wasy as `check` command)
- replicating the `config` command under `server config` so that I could better see which files are contributing to the configs

Anyway, dang robot decided to commit and create a PR. I did not intend to do that yet.

I've converted it to a draft but am more than happy to close it if you prefer. Please just let me know! I still would like to work on these two things but wanted to better communicate with you before _actually_ pushing up changes for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **CLI: `rumdl server config` command for LSP configuration inspection**
+  - New `rumdl server config` command provides per-file configuration resolution
+  - Subcommands: `file` (show config file path), `get <key>` (query specific settings)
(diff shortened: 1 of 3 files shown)


rvben:
Hi @elementalvoid,

No worries, been there! I've looked at the commits, and the first commit is a fix for a bug that was present, so I cherrypicked it. For the second effort, I have to think about what we really need in that area. Typically it's easier to keep a PR like this small to be able to see if we want to move in the proposed direction.

elementalvoid:
👍 I'll go ahead and close this then. The fallback fix was the main thing I wanted to send your way. The rest, I was just messing about with.

I don't personally _need_ the server config option and I wasn't yet happy with the implementation. If you feel like it might be useful, I'm happy to move the conversation to an issue to discuss further.

PR #145, opened 2025-11-10
perf(rules): process rules in parallel per file

ion098:
This needs profiling to determine whether it's faster for:
* Checking a single file
* Checking multiple files

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,6 +25,7 @@ pub use rules::*;
 pub use crate::lint_context::{LineInfo, LintContext, ListItemInfo};
 use crate::rule::{LintResult, Rule, RuleCategory};
 use std::time::Instant;
+use rayon::prelude::*;
 
 /// Content characteristics for efficient rule filtering
 #[derive(Debug, Default)]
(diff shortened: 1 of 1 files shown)


rvben:
Hi @ion098,

Thanks for raising this PR. I have looked at processing rules in parallel, but in my testing this did not lead to better performance, it just makes things more complex. Did you have different results?

ion098:
re @rvben 

> I have looked at processing rules in parallel, but in my testing this did not lead to better performance, it just makes things more complex. Did you have different results?

I haven't actually tested the performance yet, hence the draft PR. I wouldn't expect this to increase performance for multiple files where all cores are likely saturated, but it might for just checking a single file (though this depends on the time taken by each rule vs rayon overhead). Did you test performance with a single large file?

rvben:
What I've seen is that the most time is spent on parsing and preparing the data for the linting rules, the execution of the rules themselves doesn't take much time. In the end I had to conclude running rules in parallel doesn't improve performance, while it increases complexity. Running on files in parallel was a different story.

rvben:
The parsing is up-front and that is done in parallel. Running the rules themselves in parallel is unpredictable, more complex, and takes more time than running them serially. I am really open to performance improvements, so if you see places where it makes sense and the tests show a performance gain, let me know.

PR #180, opened 2025-12-04
Add aur packages information

BarbUk:

PR #181, opened 2025-12-04
Add archlinux and termux section

BarbUk:
Following https://github.com/rvben/rumdl/pull/180, I see that I forgot the section in the TOC.

The section for termux is also missing, so I added it too.


rvben:
Good catch, thanks!

PR #230, opened 2025-12-25
docs: use up-to-date instructions for nvim

chrisgrieser:
the instructions for nvim are a bit outdated, nowadays it is recommended (and much easier) to set up an LSP via `vim.lsp.config`.


rvben:
Thanks @chrisgrieser! Merging it in.

PR #231, opened 2025-12-25
docs: minor fixes

chrisgrieser:
- add missing reference to LSP docs in main README
- cleanup the ToC, that is, remove the preceding headings and the ToC itself from the ToC.


rvben:
Looks good, @chrisgrieser. Thanks!

PR #244, opened 2025-12-29
docs: add mise as an installation option

kvokka:
Added a section for using mise for dependency management.

Link to the [PR in mise repo](https://github.com/jdx/mise/pull/7494)


rvben:
Thanks @kvokka, great to see mise is now also a supported installation method. 👍

PR #315, opened 2026-01-20
Fix MD051 location in tables

Ravlen:
Just a quick docs fix as I was surprised MD051 ran when I was testing list-related rules. MD051 is related to _links_, not lists, so should be in the table with link-related rules.


rvben:
Thanks @Ravlen for catching this and fixing this. Merged! 🚀

PR #333, opened 2026-01-22
Add completions subcommand to generate shell completions

tpoliaw:
Supports the shells handled by clap_complete - currently zsh, bash,
fish, powershell and elvish.

How to use the generated scripts depends on the shell being used:
* Zsh - Write script to a file called _rumdl in a directory included on
  the fpath
* Bash - Source the generated script, eg by adding
  `source <(rumdl completions bash)` at the end of ~/.bashrc
* Fish - Write the script to file called rumdl.fish in a directory
  included in the $fish_complete_path variable, eg
  ~/.config/fish/completions
* PowerShell - Refer to power shell documentation
* Elvish - Refer to elvish documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,6 +360,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.5.65"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "430b4dc2b5e3861848de79627b2bedc9f3342c7da5173a14eaa5d0f8dc18ae5d"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.5.49"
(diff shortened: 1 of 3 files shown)


tpoliaw:
I've never used powershell or elvish but will try and figure out how their completions work. Should the CLI help provide instructions in the help text for each shell or is it sufficient to say "Refer to shell documentation for usage instructions" or similar?

tpoliaw:
It looks like the PowerShell version should be, write to a file then use `Import-Module C:\path\to\file` in the user's profile

rvben:
Thanks for the PR! The implementation looks good.

For the help text question: I'd keep the CLI help minimal - just "Generate shell completion scripts" is fine. Users who know they want completions generally know how to install them for their shell, and the details vary by OS/setup anyway.

The usage instructions in the PR description are helpful though - we can add those to the README or docs later.

Feel free to mark it ready for review when you're happy with it.

PR #341, opened 2026-01-27
test: add profile focused on smoke tests

kachick:
The recently added property-based tests are effective for finding edge cases, but they have introduced some flakiness. Currently, the `quick` and `ci` profiles do not exclude these randomized tests.

This PR introduces a `smoke` profile to provide a lightweight and stable test suite. It excludes heavy, randomized tests to ensure consistent results, which is useful for [packaging in Nixpkgs](https://github.com/NixOS/nixpkgs/pull/484186#pullrequestreview-3708797643).

This allows us to benefit from proptest in specific contexts while maintaining a stable path for packaging and smoke testing.

PR #343, opened 2026-01-27
fix(md037): return absolute ranges in check

kachick:
When addressing https://github.com/rvben/rumdl/pull/341 and seeing [CI result](https://github.com/rvben/rumdl/actions/runs/21373231101/job/61522910329), the proptest for MD037 often fails.
It seems to come from the difference between relative and absolute positions. How about this change?


rvben:
Thanks @kachick! Great catch on the relative vs absolute range issue.

kachick:
Thank you!

PR #371, opened 2026-02-04
docs(installation): update instructions for Arch Linux

orhun:
rumdl is now an official package: https://archlinux.org/packages/extra/x86_64/rumdl/ 🥳


rvben:
Great stuff! Thanks @orhun ! 🚀

PR #419, opened 2026-02-12
fix(docs): remove line that splits table in rules documentation in two

eread:
Remove line that splits table in rules documentation in two

https://github.com/rvben/rumdl/blob/94fa0f9e97532c48e5b69892bfb514de8df5a0c6/docs/md046.md?plain=1#L92

### DIFF
--- a/docs/md046.md
+++ b/docs/md046.md
@@ -89,7 +89,6 @@ style = "consistent"  # Options: "consistent", "fenced", "indented"
 | Value        | Description                                       |
 | ------------ | ------------------------------------------------- |
 | `consistent` | All code blocks must use the same style (default) |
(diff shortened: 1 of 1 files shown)


eread:
@rvben Could you review this one? 🙇🏻

pygarap:
@rvben PRs like this set a bad precedent for `rumdl`, because they fix the Markdown docs instead of fixing `rumdl` itself. The correct fix is to update the Rust code so `rumdl` can auto-fix the Markdown violations into the correct format.
That is why I opened issue #420 to track it.

rvben:
Oops! Good find. Thanks for the PR, @eread.

PR #421, opened 2026-02-12
fix(MD054): match message with rule configuration option

eread:
Match message with rule MD054  configuration option.

Currently, the message that's returned doesn't 100% match the configuration option itself.

### DIFF
--- a/src/config/registry.rs
+++ b/src/config/registry.rs
@@ -228,6 +228,7 @@ pub static RULE_ALIAS_MAP: phf::Map<&'static str, &'static str> = phf::phf_map! 
     "MD072" => "MD072",
     "MD073" => "MD073",
     "MD074" => "MD074",
+    "MD075" => "MD075",
 
     // Aliases (hyphen format)
     "HEADING-INCREMENT" => "MD001",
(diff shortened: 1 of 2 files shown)


eread:
@rvben could you review this one? 🙇🏻

eread:
> Good catch. The warning message should match the config key `url-inline` rather than the internal field name `url_inline`.
> 
> Right now the PR only updates the test assertion, so the test would fail since the implementation still outputs `url_inline`. A few more spots need updating in `src/rules/md054_link_image_style.rs`:
> 
> * Line 98 in `is_style_allowed`:
>   ```rust
>   "url-inline" => self.config.url_inline,
>   ```
> * Line 208 where the match style is assigned:
>   ```rust
>   style: if text == url { "url-inline" } else { "inline" },
>   ```
> * Line 405 in the test file has a second assertion that needs the same change:
>   ```rust
>   assert!(result[0].message.contains("'url-inline'"));
>   ```
> 
> Could you update the PR with these?

@rvben I can! I'm afraid I only figured out how to run the test suite **after** I pushed this and asked for review 😅 

The tests pass now 🙇🏻

eread:
@rvben I've rebased this one against upstream `main`, and also added a commit to fix some broken tests in upstream `main`.

Could you re-review?

rvben:
Thanks @eread! Good catch on the style name mismatch. Merging.

eread:
> Thanks @eread! Good catch on the style name mismatch. Merging.

@rvben Welcome!

PR #442, opened 2026-02-20
fix(docs): remove empty configuration block from rule MD023 page

eread:
The documentation page for rule MD023 has a configuration section that doesn't have any configuration in it (apart from a TOML section heading).


eread:
@rvben Could you review? There is failing linting, but it's unrelated to this PR.

rvben:
Thanks @eread, good catch!

PR #465, opened 2026-02-25
docs: fix multi-paragraph list item indentation

sisp:
I've increased the indentation for multi-paragraph list items from 3 to 4 spaces to fix MkDocs-rendered output.

| Before | After |
| - | - |
| before | after |

### DIFF
--- a/docs/global-settings.md
+++ b/docs/global-settings.md
@@ -491,51 +491,51 @@ Result:
 
 1. **MkDocs/mdBook Projects**:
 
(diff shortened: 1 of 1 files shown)


rvben:
Thanks @sisp for pointing me here. These are always a bit of a dilemma, because I'd like rumdl to actually detect these types of issues and be able to fix them too. Resolving them manually is just 1 part. I'll make sure rumdl will be able to fix these by itself too.

PR #470, opened 2026-02-27
docs(md046): Expand documentation for rule

eread:
Document that `rumdl` can find and fix fenced code blocks that aren't closed.

### DIFF
--- a/docs/md046.md
+++ b/docs/md046.md
@@ -4,7 +4,10 @@ Aliases: `code-block-style`
 
 ## What this rule does
 
-Ensures all code blocks in your document use the same style - either fenced (with backticks) or indented (with spaces).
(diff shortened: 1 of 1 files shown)


eread:
@rvben Could you review this one? I noticed that MD046 is able to find (https://github.com/eread/rumdl/blob/440f587c1cfa45f1ea7fbe882eac97ea18f292a4/src/rules/md046_code_block_style.rs?plain=1#L712) and fix (https://github.com/eread/rumdl/blob/440f587c1cfa45f1ea7fbe882eac97ea18f292a4/src/rules/md046_code_block_style.rs?plain=1#L1003) fenced code blocks that aren't closed, but I couldn't find this written down anywhere.

WDYT?

PR #554, opened 2026-04-01
docs: direct Sublime Text users to LSP for Sublime Text documentation

eread:
With https://github.com/sublimelsp/LSP/pull/2842 merged, the documentation at https://lsp.sublimetext.io can serve as the source of truth for configuring Sublime Text with rumdl as a language server.

### DIFF
--- a/docs/lsp.md
+++ b/docs/lsp.md
@@ -120,21 +120,10 @@ Add to your Zed settings:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rvben:
Thanks @eread, makes sense to defer to the upstream docs now that rumdl is listed there. Less to maintain on our end.

PR #555, opened 2026-04-01
docs: Use `--silent` instead of `--quiet` to avoid diagnostic lines being inserted 

joelostblom:
When I run with `--quiet` I get diagnostic lines like these inserted into my buffer under the visual selection:

```
<stdin>:2:1: [MD013] Line length 269 exceeds 72 characters [fixed]
<stdin>:2:1: [MD041] First line in file should be a level 1 heading
```

Using `--silent` instead just reformats the selection without inserting such diagnostic lines (as expected).

### DIFF
--- a/docs/usage/editors.md
+++ b/docs/usage/editors.md
@@ -74,10 +74,10 @@ let g:ale_fixers = {'markdown': ['rumdl']}
 
 ```vim
 " Format selection
-:'<,'>!rumdl fmt - --quiet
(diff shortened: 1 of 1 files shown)


rvben:
Thanks @joelostblom, good catch. `--quiet` only suppresses the summary while `--silent` suppresses all non-content output, so this is the right flag for pipe-based editor integrations.

PR #561, opened 2026-04-04
fix: edit_uri should be main not master

vEnhance:
The old mkdocs defaults to master and zensical is still respecting that, so right now on rumdl.dev none of the GitHub action links work

cf https://github.com/zensical/zensical/issues/321 and https://github.com/zensical/backlog/issues/93


rvben:
Thanks @vEnhance for catching this and sending the fix! 👍

PR #603, opened 2026-05-01
fix(MD013): preserve nested tables during list reflow

### LINKED ISSUE #598: MD013 Reflow Collapses Tables Nested In List Items
### rumdl version

0.1.84

### Installation method

cargo install

### Bug description

MD013 semantic reflow can treat table rows nested under a list item as paragraph text. The fixer can collapse the table header, delimiter, and body rows into one list paragraph, which changes the rendered content from a table into plain text.

### Steps to reproduce

1. Create `test.md`:

   ```markdown
   - Error mappings should remain tabular when nested under a list item and the introductory sentence is intentionally long enough to trigger semantic reflow.
     | Code | Meaning |
     |---|---|
     | `LIMIT_REACHED` | The configured quota was reached. |
     | `TOKEN_EXPIRED` | The session token expired. |
   ```

2. Use MD013 reflow with semantic line breaks and a line length around 80.
3. Run:

   ```bash
   rumdl check --fix test.md
   ```

### Expected behavior

The nested table rows remain separate table rows.

### Actual behavior

The nested table rows can be merged into list paragraph text, breaking table rendering.

### Configuration

```toml
[MD013]
line-length = 80
reflow = true
reflow-mode = "semantic-line-breaks"
```

### Operating system

Linux

### Additional context

_No response_

mattiasgronlund:
## Description

Fixes MD013 semantic reflow so potential table rows nested inside list items are treated as structural lines, like fenced code markers. This prevents the fixer from merging nested table rows into paragraph text.

The regression test and bug fix were created in cooperation with Codex.

Fixes: #598

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests added/updated
- [ ] Manual testing performed

Added regression test:

- `test_reflow_preserves_tables_nested_in_list_items`

Validated locally:

```bash
make test-dev
make fmt
make lint
```

## Checklist

- [ ] Documentation updated (if needed)
- [ ] CHANGELOG.md updated (if user-facing)

### DIFF
--- a/src/rules/md013_line_length/mod.rs
+++ b/src/rules/md013_line_length/mod.rs
@@ -1688,9 +1688,8 @@ impl MD013LineLength {
                             if line_info.is_div_marker {
                                 list_item_lines.push(LineType::DivMarker(content));
                             }
-                            // Check if this is a fence marker (opening or closing)
-                            // These should be treated as code block lines, not paragraph content
-                            else if is_fence_marker(&content) {
+                            // Check for structural lines that must not be merged into paragraphs.
+                            else if is_fence_marker(&content) || TableUtils::is_potential_table_row(&content) {
                                 list_item_lines.push(LineType::CodeBlock(content, indent));
                             }
                             // Check if this is a semantic line (NOTE:, WARNING:, etc.)
--- a/src/rules/md013_line_length/mod.rs
+++ b/src/rules/md013_line_length/mod.rs
@@ -1689,7 +1689,7 @@ impl MD013LineLength {
                                 list_item_lines.push(LineType::DivMarker(content));
                             }
                             // Check for structural lines that must not be merged into paragraphs.
-                            else if is_fence_marker(&content) || TableUtils::is_potential_table_row(&content) {
+                            else if is_fence_marker(&content) {
                                 list_item_lines.push(LineType::CodeBlock(content, indent));
                             }
                             // Check if this is a semantic line (NOTE:, WARNING:, etc.)
--- a/tests/rules/md013_test.rs
+++ b/tests/rules/md013_test.rs
@@ -1004,6 +1004,28 @@ fn test_reflow_preserves_tables_nested_in_list_items() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


rvben:
Thanks for digging into this. Two things I'd like to flag before merging:

**1. The regression test doesn't catch the bug.**

I reverted only the `src/rules/md013_line_length/mod.rs` change on this branch and re-ran `test_reflow_preserves_tables_nested_in_list_items` — it still passes. Running the issue's exact reproducer through `rumdl check --fix` on current `main` (without this PR) also preserves the table rows.

A different input does reproduce the underlying bug on `main`:

```markdown
- A long sentence about something interesting that describes how pipes appear in command-line examples.
  cell a | cell b
```

On `main`, `cell a | cell b` gets merged into the previous sentence:

```
- A long sentence about something interesting
  that describes how pipes appear in command-line examples. cell a | cell b
```

With this PR, it stays on its own line. So the fix is real, but the current test is a smoke test against a case that already passes on `main`.

Per the project's testing philosophy, could you replace or supplement the test with a reproducer that genuinely fails before the fix?

**2. Fix placement.**

The change is at line 1693 in the `is_fence_marker` branch, which fires before the careful table-detection at lines 1718–1727. After this PR, any `TableUtils::is_potential_table_row(&content)` matches first and routes everything through `LineType::CodeBlock`, so the `LineType::Table` classification path becomes unreachable for indented table content under list items.

The `Table` variant exists for a reason (downstream formatting and table-aware behaviour). I'd suggest broadening the existing branch at line 1718 instead — extend its predicate to also catch orphan pipe-shaped lines (e.g. drop the `pipe_bordered || next_is_delim || prev_was_table` requirement when there's clearly enough pipe structure), so the result is still classified as `Table`. That keeps `LineType` semantics consistent.

Happy to chat through the design if helpful.

mattiasgronlund:
```markdown
- A long sentence about something interesting that describes how pipes appear in command-line examples.
  cell a | cell b
```
Is not a nested table, I think this issue was actually fixed in 0.1.85, but I will try to understand the above issue, and provide a new pull request for that one.

PR #604, opened 2026-05-01
fix(MD013): preserve wildcard asterisks during reflow

### LINKED ISSUE #597: MD013 Reflow Inserts Space Before Literal Wildcard Asterisks
### rumdl version

0.1.84

### Installation method

cargo install

### Bug description

MD013 semantic reflow can insert a space before a literal wildcard asterisk in identifiers such as `TICKET-*`. This changes literal text, e.g. `TICKET-*` becomes `TICKET- *`.

### Steps to reproduce

1. Create `test.md`:

   ```markdown
   - **Catalog layout.** Reference files live under `docs/items/<name>/` (`README.md`, `notes.md`, `checks.md`) and `examples/items/<name>.md`. IDs (TICKET-*, TASK-*, CASE-*) stay in the shared namespace: TICKET-17 is still called TICKET-17 inside the nested catalog.
   ```

2. Use MD013 reflow with semantic line breaks and a line length around 80.
3. Run:

   ```bash
   rumdl check --fix test.md
   ```

### Expected behavior

Literal wildcard identifiers remain unchanged, e.g. `TICKET-*`.

### Actual behavior

The fixer can insert whitespace before the asterisk, e.g. `TICKET- *`.

### Configuration

```toml
[MD013]
line-length = 80
reflow = true
reflow-mode = "semantic-line-breaks"
```

### Operating system

Linux

### Additional context

_No response_

mattiasgronlund:
## Description

Fixes MD013 semantic reflow so the emphasis sentence-split path does not insert a space after a hyphen before a literal wildcard asterisk. This preserves identifiers like `TICKET-*` during reflow.

The regression test and bug fix were created in cooperation with Codex.

Fixes: #597

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests added/updated
- [ ] Manual testing performed

Added regression test:

- `test_reflow_preserves_literal_wildcard_asterisks`

Validated locally:

```bash
make test-dev
make fmt
make lint
```

## Checklist

- [ ] Documentation updated (if needed)
- [ ] CHANGELOG.md updated (if user-facing)

### DIFF
--- a/src/utils/text_reflow.rs
+++ b/src/utils/text_reflow.rs
@@ -1350,6 +1350,14 @@ fn parse_markdown_elements_inner(text: &str, attr_lists: bool) -> Vec<Element> {
     elements
 }
 
+fn should_insert_space_before_join(current: &str) -> bool {
+    !current.is_empty()
+        && !current.ends_with(' ')
+        && !current.ends_with('(')
+        && !current.ends_with('[')
+        && !current.ends_with('-')
+}
+
 /// Reflow elements for sentence-per-line mode
 fn reflow_elements_sentence_per_line(
     elements: &[Element],
@@ -1471,12 +1479,7 @@ fn reflow_elements_sentence_per_line(
             };
 
             // Add space before element if needed, but not for adjacent elements
-            if !is_adjacent
-                && !current_line.is_empty()
-                && !current_line.ends_with(' ')
-                && !current_line.ends_with('(')
-                && !current_line.ends_with('[')
-            {
+            if !is_adjacent && should_insert_space_before_join(&current_line) {
                 current_line.push(' ');
             }
             current_line.push_str(&element_str);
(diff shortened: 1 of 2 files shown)


rvben:
Thanks for the fix — the wildcard case is real and reproduces on `main`.

One concern before merging: the change patches two of three sibling sites in `src/utils/text_reflow.rs`. The same pattern at line 1477 (the general element-joining path) doesn't get the new `!current_line.ends_with('-')` check, so a hyphen-attached emphasis going through that path will reintroduce the bug.

Could you either:

- (a) Apply the same `&& !current_line.ends_with('-')` at line 1477, or
- (b) Extract a shared helper, e.g.:

  ```rust
  fn should_insert_space_before_join(current: &str) -> bool {
      !current.is_empty()
          && !current.ends_with(' ')
          && !current.ends_with('(')
          && !current.ends_with('[')
          && !current.ends_with('-')
  }
  ```

  and call it at all three sites?

I tested `dash-*emphasis*` (no space) on this branch and it now produces the correct output, so the fix itself is sound — just want to keep the three sibling conditionals consistent.

Also worth noting: the fix actually addresses a broader class than wildcards specifically — any hyphen-attached emphasis-like construct. The PR title and test could mention that, but not a blocker.

mattiasgronlund:
I fixed the code according to your comment, using fn should_insert_space_before_join(current: &str) -> bool.

I tried to create a test case which would fail unless I added the fix on line 1477 also, but I had to give up.
I am quite sure that if is_adjacent is false, current_line can not end with '-'.

PR #616, opened 2026-05-06
Warm cache profiling path

### LINKED ISSUE #613: Optimize and profile the warm-cache check path
### Problem or motivation

Warm-cache runs should be very fast, but it is hard to see where time is spent and some hot-path work is still repeated even when cached results are available.

Areas that can affect warm-cache performance include:

- repeated file content hashing
- cache entry read/parse/validation overhead
- cache write serialization and I/O
- unnecessary mutex contention around cache stats/lookups
- repeated config/rule hash work
- discovery overhead before linting
- profiling timers that depend on shared global start/stop state
- building file indexes and running cross-file checks around cached single-file warnings

Without finer profiling and cache-path optimization, regressions in warm-cache performance are hard to diagnose.

### Proposed solution

Add a broader warm-cache optimization/profiling path.

Implementation shape from the branch:

- Make cache stats internally synchronized so cache methods can take `&self` instead of requiring broader mutable locking.
- Support precomputed file hashes for cache get/set operations to avoid hashing content more than necessary.
- Add reason-preserving cache lookup APIs for diagnostics.
- Instrument cache hashing, cache read, parse, serialize, and write sections under the profiling feature.
- Refactor processing/check-run paths so warm-cache operations avoid unnecessary work where possible.
- Improve profiling reporting to describe total measured time, noting that parallel sections may overlap.
- Change scoped profiling timers to record elapsed duration directly rather than relying on shared active-timer start/stop bookkeeping.
- Include discovery optimizations such as compiled exclude matchers and walker file-type checks.

### Alternatives considered

Only add more profiling labels without changing the cache path. That would improve visibility but would not address the repeated work visible in warm-cache runs.

### Feature category

Performance

### Example markdown

```markdown

```

### Contribution

### Additional context

_No response_

mattiasgronlund:
Fixes #613

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -211,7 +211,7 @@ impl LintCache {
             return Err(CacheMissReason::Disabled);
         }
 
-        let cache_path = self.cache_file_path(&file_hash, rules_hash);
+        let cache_path = self.cache_file_path(file_hash, rules_hash);
 
         // Try to read cache file
         #[cfg(feature = "profiling")]
@@ -294,7 +294,7 @@ impl LintCache {
             return;
         }
 
-        let cache_path = self.cache_file_path(&file_hash, rules_hash);
+        let cache_path = self.cache_file_path(file_hash, rules_hash);
 
         // Create cache directory if it doesn't exist
         if let Some(parent) = cache_path.parent() {
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -10,10 +10,33 @@
 use rumdl_lib::rule::LintWarning;
 use serde::{Deserialize, Serialize};
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Mutex;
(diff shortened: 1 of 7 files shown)


rvben:
Merged as 5c2a00b0 — thanks @mattiasgronlund for the optimization work. Two follow-ups went in on top before merge:

1. **Atomic cache writes** (ccb7ee18). Removing the `Mutex` left a real corruption window: parallel workers can produce identical cache keys for distinct files (same content + config + rules), and `CacheEntry.timestamp` differs per write. Switched to a tempfile + `fs::rename` so concurrent writers can't interleave bytes. Added `test_atomic_write_concurrent_no_corruption` (16 threads × 32 iterations) to lock the invariant in.

2. **Feature-gated profiling macros** (0068cae7). `time_section!` and `time_function!` were unconditionally calling `section.to_string()`, so default builds (without the `profiling` feature) paid a `String` allocation per call site on the hot path. Now expand to `$block` when the feature is off.

Will ship in the next release.

PR #617, opened 2026-05-06
Optimize markdown discovery filters

### LINKED ISSUE #612: Optimize markdown file discovery filters
### Problem or motivation

# Problem or motivation
File discovery can do extra repeated work when applying exclude patterns, especially in repositories with many files or several explicit/configured excludes.

Current behavior repeatedly compiles glob patterns while checking explicit files and during post-walk filtering. It also uses `path.is_file()` during walker iteration, which can require extra filesystem metadata lookups even though the walker already has file type information.

This adds avoidable overhead to `rumdl check .`, particularly in large repositories and pre-commit workflows.

### Proposed solution

Optimize discovery by compiling exclude matchers once and reusing them throughout discovery.

### Alternatives considered

Leave the logic as-is because the behavior is already correct. That keeps code simpler, but preserves repeated glob compilation and filesystem checks in hot discovery paths.

### Feature category

Performance

### Example markdown

```markdown

```

### Contribution

### Additional context

_No response_

mattiasgronlund:
Fixes #612

### DIFF
--- a/src/file_processor/discovery.rs
+++ b/src/file_processor/discovery.rs
@@ -1,6 +1,7 @@
 //! File discovery, path utilities, and pattern expansion
 
 use core::error::Error;
+use globset::{Glob, GlobMatcher};
 use ignore::WalkBuilder;
 use ignore::overrides::OverrideBuilder;
 use rumdl_config::resolve_rule_names;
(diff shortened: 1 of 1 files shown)


rvben:
Merged as 49a1de2c — thanks @mattiasgronlund. Precompiling the exclude matchers once and reusing the walker `Types` is exactly the right shape for the hot path. Will ship in the next release.

PR #619, opened 2026-05-06
Allow profile output in silent mode

### LINKED ISSUE #615: --silent overrides --profile
### rumdl version

0.1.89

### Installation method

Built from source

### Bug description

Currently `--silent` override `--profile`.

`--profile` is an explicit output request. If a user passes `--profile`, the profiling report should still be emitted even when `--silent` is also present.

### Steps to reproduce

cargo build  --release --features profiling
target/release/rumdl check --profile --silent

### Expected behavior

Profile output should be printed

### Actual behavior

 profile output is not printed.

### Configuration

```toml

```

### Operating system

Linux

### Additional context

_No response_

mattiasgronlund:
Fixes #615

### DIFF
--- a/src/check_runner.rs
+++ b/src/check_runner.rs
@@ -566,14 +566,23 @@ pub fn perform_check_run(ctx: &CheckRunContext<'_>) -> (bool, bool, bool, usize)
         formatter::print_statistics(&all_warnings_for_stats);
     }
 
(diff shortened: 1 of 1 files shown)


rvben:
Merged as d9cde4d6 — thanks @mattiasgronlund. Clean fix, and good call adding the test that pins the exact behavior. Will ship in the next release.

PR #623, opened 2026-05-08
docs: update rumdl version referred in readme

mattcristal:
## What was done

Ideally, the version of `rumdl` referred in the README should be the same as the latest version of the package.

This makes it easier, for example, for users to select the latest pre-commit version.

It would be nice to find a way to automate this.


rvben:
Thanks @mattcristal, this actually was automated before. Seems that it started to miss a couple. Thanks! Fixed it now.

PR #626, opened 2026-05-09
refactor: make rules and folder structure organization more omogeneous

mattcristal:
### Key changes

While the vast majority of the rules followed the modern Rust 2018+ module pattern (having a `rule_name.rs` file alongside a `rule_name/` directory containing submodules like `config.rs`), five specific rules were still using the older `mod.rs` pattern (`rule_name/mod.rs`). 

I have standardized the directory to make it fully consistent by converting the following files to the modern structure:
- `md013_line_length/mod.rs` → `md013_line_length.rs`
- `md041_first_line_heading/mod.rs` → `md041_first_line_heading.rs`
- `md060_table_format/mod.rs` → `md060_table_format.rs`
- `md061_forbidden_terms/mod.rs` → `md061_forbidden_terms.rs`
- `md063_heading_capitalization/mod.rs` → `md063_heading_capitalization.rs`

In addition to the key changes for why I opened this PR, I also had to make some other changes to make the pre-commit work.

#### Pre-commits config

For example:

Here:

https://github.com/rvben/rumdl/blob/main/CONTRIBUTING.md

Suggests to use pre-commits (and not prek). So, I had to change the pre-commit file, specifically remove: - repo: builtin, otherwise it would not work.

#### actionlint

I could also not find in the `.github` folder the `actionlint` workflow.


mattcristal:
@rvben my personal suggestion is to review this PR and, if okay, merge it as-is. 

Then, we can open another PR to transition from `pre-commit` -> `prek` 

That, however, will also require to change the dependencies, see the last one here:

https://github.com/rvben/rumdl/blob/main/pyproject.toml

.mise.toml
```diff
@@ -38,9 +40,6 @@ zig = "0.13"
 # separately. maturin is already installed via cargo above.
 
 [settings]
-# Use cargo-binstall when available for faster installations
-cargo_binstall = true
```
rvben, line 42: Why remove this?

rvben:
Hi @mattcristal, thanks for this PR. I feel we should update the CONTRIBUTING.md to match reality (prek) instead of matching CONTRIBUTING.md by changing back from prek to pre-commit. The other changes look good to me.

mattcristal:
@rvben let me do some updates

.mise.toml
mattcristal: Because I am getting:

```shell
  mise::config::parse_error

    × Invalid TOML in config file: /Users/304079/Documents/Matteo/
    │ rumdl/.mise.toml
      ╭─[~/Documents/Matteo/rumdl/.mise.toml:43:18]
   42 │ # Use cargo-binstall when available for faster installations
   43 │ cargo_binstall = true
      ·                  ──┬─
      ·                    ╰── invalid type: boolean `true`, expected tool list
   44 │ 
      ╰────
```

rvben:
Merged! 🚀 Thanks @mattcristal for the PR. 💪

PR #645, opened 2026-05-29
Add reference to 'sentence-per-line' mode description

cbachhuber:
I found this nice reference that lists the surpisingly many advantages of one sentence per line.


rvben:
Thanks @cbachhuber, merged! That helps clarify it! 🚀

PR #652, opened 2026-06-02
fix(MD049,MD037): Preserve inline code inside emphasis during auto-fix

### LINKED ISSUE #651: bug(MD049/MD037): rumdl fmt replaces inline code inside emphasis with X placeholders
### rumdl version

0.2.5 (also reproduces on `main`)

### Installation method

mise

### Bug description

When `rumdl fmt` auto-fixes an emphasis span that contains an inline-code span, the inline code, including its surrounding backticks, is replaced in the written file by a run of literal `X` characters of equal length. The original code content is permanently lost, so this corrupts the document.

Internally, rumdl masks inline code with equal-length `X` runs so that backtick contents are not mistaken for emphasis markers during detection. That masked copy is meant to be used only for detection, but for these two rules the masked text leaks into the fix replacement instead of the original text being restored.

Affected rules:

- MD049 (emphasis style, `_` to `*` or `*` to `_`)
- MD037 (spaces inside emphasis)

MD050 (strong style) is not affected.

### Steps to reproduce

1. Create `test.md`:

   ```markdown
   - _An item with `inline code` inside._
   ```

2. Create `.rumdl.toml`:

   ```toml
   [MD049]
   style = "asterisk"
   ```

3. Run:

   ```bash
   rumdl fmt test.md
   ```

### Expected behavior

The inline code is preserved verbatim and only the emphasis markers change:

```markdown
- *An item with `inline code` inside.*
```

### Actual behavior

The inline code is replaced by `X` placeholders (here `` `inline code` `` is 13 characters, so 13 `X`s):

```markdown
- *An item with XXXXXXXXXXXXX inside.*
```

The same corruption occurs for MD037. Given the input `A paragraph with * spaced `inline code` emphasis * here.`, `rumdl fmt` produces `A paragraph with *spaced XXXXXXXXXXXXX emphasis* here.`

### Configuration

```toml
[MD049]
style = "asterisk"
```

### Operating system

macOS

### Additional context

- This is `fmt`-only. `rumdl check` does not modify files, so it does not corrupt content (it only reports the warning).
- Independent of line length/MD013 reflow, list vs. paragraph context, and the emphasis direction; the only requirement is that an inline-code span sits inside an emphasis span being fixed.
- Root cause: the fix replacement is built from the code-masked copy of the line (`replace_inline_code`) rather than the original text.
- Likely origin: For MD049, commit `1d38d7d1` (first released in `v0.0.90`) replaced the earlier AST-based emphasis fix, which read the replacement text straight from the original document, with the current approach that detects emphasis on the `X`-masked copy of the line and then reuses that masked text in the fix. For MD037, commit `bac52d8a` (first released in `v0.0.72`) made the equivalent change, switching from a regex over the original line to detection on the masked copy whose content then feeds the fix. In both, masking was added for detection but the fix was never switched back to the original text.

thegeorgeliu:
## Description

To detect emphasis markers reliably, `MD049` and `MD037` first build a throwaway copy of each line with every inline code span replaced by an equal-length sequence of `X` characters (and inline math by `M`). This prevents backtick or dollar contents from being misread as emphasis markers. The auto-fix, however, reused the emphasis text taken from that masked copy, so the `X`/`M` placeholders leaked into the written file and permanently replaced the original inline code.

The fix takes the replacement text from the original line at the span's byte offsets instead. Because masking replaces each span with a sequence of the same byte length, those offsets remain valid against the original line.

Example before this fix (`rumdl fmt` with MD049 `style = "asterisk"`):

- input: `- _An item with `inline code` inside._`
- output: `- *An item with XXXXXXXXXXXXX inside.*`

After the fix the inline code is preserved: `- *An item with `inline code` inside.*`. Only `fmt`/`--fix` was affected; `check` does not modify files. `MD050` (strong style) was not affected.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests added/updated (regression tests in both rules)
- [x] All tests passing: `cargo nextest` for md049/md037/emphasis/reflow/md013/formatter_idempotency, and full `cargo test --lib` (5810 passed)
- [x] Manual testing performed: `rumdl fmt` on the repro is now lossless and idempotent

## Checklist

- [x] Code follows project style (`cargo fmt --check` clean; CI clippy gate `-D warnings -D clippy::uninlined_format_args` clean)
- [ ] Documentation updated (not needed)
- [ ] CHANGELOG.md updated (maintainer-generated at release via git-cliff)

Closes #651

### DIFF
--- a/src/rules/md037_spaces_around_emphasis.rs
+++ b/src/rules/md037_spaces_around_emphasis.rs
@@ -319,8 +319,14 @@ impl MD037NoSpaceInEmphasis {
                     format!("{marker_char}{marker_char}")
                 };
 
-                // Create the fixed version by trimming spaces from content
-                let trimmed_content = span.content.trim();
+                // Create the fixed version by trimming spaces from content.
+                // Slice the content from the *original* line, not from the
+                // code/math-masked copy: `span.content` would contain the 'X'/'M'
+                // placeholders, which must never leak into the generated fix.
+                // Masking is length-preserving, so the span byte offsets are
+                // valid in `content`.
+                let original_content = &content[span.opening.end_pos()..span.closing.start_pos];
+                let trimmed_content = original_content.trim();
                 let fixed_text = format!("{marker_str}{trimmed_content}{marker_str}");
 
                 // Truncate long emphasis spans for readable warning messages
@@ -398,6 +404,19 @@ mod tests {
             !result.is_empty(),
             "Expected warnings for spaces in emphasis outside code block"
         );
+    }
+
+    #[test]
+    fn test_inline_code_inside_spaced_emphasis_preserved_on_fix() {
+        // Regression test: inline code inside a spaced emphasis span must survive the
+        // space-trimming fix. Previously the 'X' masking placeholder leaked into
+        // the fix output, destroying the code span.
+        let rule = MD037NoSpaceInEmphasis;
+        let content = "Set * the `id` field * below.";
+        let ctx = LintContext::new(content, crate::config::MarkdownFlavor::Standard, None);
+        let fixed = rule.fix(&ctx).unwrap();
+        assert_eq!(fixed, "Set *the `id` field* below.");
+        assert!(!fixed.contains('X'), "masking placeholder leaked into fix: {fixed}");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


rvben:
Thanks @thegeorgeliu, for both the clear bug report in #651 and the fix. Nicely scoped, with good regression tests. Merging now; it'll ship in the next patch release.

thegeorgeliu:
Happy to contribute @rvben! Thanks for all your great work on rumdl, I've been using it for a while now throughout all my personal and work projects!

Issue #644, opened 2026-05-29
More column-style ordered-list indentation across different width markers


chandlerc:
### Problem or motivation

`MD030` enforces a fixed number of spaces after the list marker via
`ol-single` / `ol-multi`. That works fine when every marker in a list has the
same width, but it produces inconsistent alignment as soon as a list crosses a
digit boundary (`9.` → `10.`, `99.` → `100.`):

```markdown
<!-- ol-single = 2, ol-multi = 2 -->
1.  one          ← text in column 4
9.  nine         ← text in column 4
10.  ten         ← text in column 5
```

Prettier, by contrast, adjusts the padding per item so the text column stays fixed in the most common cases, and this is the de-facto convention in many existing markdown files:

```markdown
1.  one          ← 2 spaces after `.`
9.  nine         ← 2 spaces after `.`
10. ten          ← 1 space after `.`
```

While it is tempting to solve this by using `1.` as the only ordered list marker, this is unfortunate for markdown files that are read both in their plain text and rendered form, as it makes it harder to reference specific list items in plain text.

### Proposed solution

A number of options for how best to address this...

1. `rumdl` could enforce _alignment_, padding lines to match whatever indent is used for the widest ordered marker.

2. `rumdl` could support width-specific indent settings, and let projects configure as many as they need -- realistically, I think handling 3 digit numbers is likely _very_ rare, and 4 digit numbers seem contrived at best. So it would be easy for projects to use this to achieve the layout they want.

For (1) specifically, I would suggest also suggest using a "tab-stop" approach to indentation rather than fixed number of spaces so that the indentation can step between well defined points rather than moving incrementally.

### Feature category

CLI / configuration

### Example markdown

```markdown

```

### Contribution

- [x] I would be willing to submit a PR to implement this feature

rvben:
Thanks @chandlerc, a PR would be very welcome. The approach is feasible: rumdl already computes `ListBlock.max_marker_width`, which is exactly what column alignment needs. Suggest an opt-in `ol-align` (default off) with `expected_spaces = (max_marker_width - 1 - marker.len()) + configured_spaces`, floored at 1. Happy to review.

chandlerc:
I've been working through a PR to address this, and ended up finding a slightly different approach working better. Wondering if you'd be OK with that. I'll also post the PR shortly, and happy to iterate either way.

Two things that made me look for a somewhat different approach:

[...]

PR #657, opened 2026-06-06
fix(md030): keep nested content attached when a marker widens

chandlerc:
MD030 fixes the spaces after a list marker, but when that fix *widens* a
multi-line item's marker indent, it only moved the content on the same line as
the marker, leaving the continuation content behind. For a nested list that
detaches it: the child marker ends up left of the parent's content column and
flattens into a sibling. With a configured `ol-multi`/`ul-multi` greater than
the current spacing, `--fix` could therefore corrupt the document, and no other
rule recovers it.

Track the open list items on a stack, each carrying its cumulative indent shift
(its own marker re-spacing plus every ancestor's), and re-indent continuation
and nested lines to follow a widened marker. The shifts accumulate across
nesting levels in a single pass. An inline nested bullet that shares the
marker's line (`- x` in `1. - x`) is spaced and shifted like a sibling bullet.

Re-indent only when content must move *right*. Narrowing leaves content
over-indented but still attached, which MD077 (list continuation
indentation) tightens, so MD030 leaves that to it. The idea is that
MD030 owns under-indent, MD077 owns over-indent.

The whole mechanism is skipped unless a configured spacing exceeds 1, so
the default configuration is unaffected.

There is a pure refactor commit first to expose needed functionality,
followed by the fix itself.

This also introduces a use of `indoc` to allow for easier construction of
multi-line test cases with visible indentation structures.

This is a first step towards fixing #644.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1316,6 +1316,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "indoc"
+version = "2.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79cf5c93f93228cf8efb3ba362535fb11199ac548a09ce117c9b1adc3030d706"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "inotify"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2293,6 +2302,7 @@ dependencies = [
  "globset",
  "ignore",
  "indexmap",
+ "indoc",
  "jsonschema",
  "libc",
  "log",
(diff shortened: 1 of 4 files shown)


chandlerc:
Let me know if you'd like these changes re-organized in any way, or if there is a better way to do this.

One thing that I did consider but ended up not doing is merging MD077's logic into MD030, so that a single pass of logic can both indent and outdent as needed. But it seemed better to not disturb the structure of the different checks.

chandlerc:
(Also, just to be clear, I'm still a Rust novice, so many apologies if any of this is totally wrong or not using the right idioms.)

rvben:
Looks good @chandlerc, nice 👌 I'll merge them in 🚀

PR #688, opened 2026-06-25
docs(pre-commit): fix version typo in suggested pre-commit config

tomjholland:
Version given in example is 0.0.222, rather than 0.2.22.


rvben:
Thanks @tomjholland, merged! 🚀

PR #689, opened 2026-06-25
fix(lint-context): attach list items whose marker line opens a code fence

chandlerc:
A list marker line that also opens a fenced code block (e.g. "- ```python" or "1. ```js") is flagged `in_code_block`. `parse_list_blocks` short-circuited at its top-of-loop `in_code_block` guard before inspecting `list_item`, so the item was never registered and everything indented under it — continuation paragraphs and nested lists — was dropped from the list model. Tilde-fenced first items produced no list block at all.

Skip the code-block branch only when the line is not also a list item, letting the marker line fall through to normal list-item handling. Lines inside a fence are never flagged as list items, so this matches only the genuine marker line; marker-like lines within the fenced content stay out of the list. Inline code spans (single- or multi-line) were never affected.

Add regression tests for the parser fix (fenced item mid-list, as the first item, with a nested list, plus tilde and ordered variants and inline-code control cases) and for the rules that lint the opening fence line (MD040, MD031, MD046), confirming they still apply when the fence opens on a marker line.

Assisted-by: Claude Code

### DIFF
--- a/src/lint_context/list_blocks.rs
+++ b/src/lint_context/list_blocks.rs
@@ -88,8 +88,16 @@ pub(super) fn parse_list_blocks(content: &str, lines: &[LineInfo]) -> Vec<ListBl
     for (line_idx, line_info) in lines.iter().enumerate() {
         let line_num = line_idx + 1;
 
-        // Enhanced code block handling using Design #3's context analysis
-        if line_info.in_code_block {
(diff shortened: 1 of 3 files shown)


rvben:
Thanks for the fix, @chandlerc ! Looks good. Merged.

PR #692, opened 2026-06-26
feat(MD013): reflow lists with configured marker spacing

chandlerc:
Before, MD013 reflow hard-coded one space after a list marker and a marker-width continuation indent. It ignored MD030's configured spacing (ul/ol single/multi and ol-align-column) and fought MD030's own fixer.

After, in the default and normalize reflow modes MD013 derives the post-marker spacing from MD030 and aligns continuation lines to the resulting content column, for top-level, nested, and blockquote-nested items. Sentence/semantic modes and the MkDocs flavor are unchanged, and with the default MD030 (a single space) the reflow output is byte-identical to before.

To avoid two implementations that can drift, the shared logic lives in one place each: MD030Config gains expected_spaces() (the single- vs multi-line and ol-align-column rules), and the continuation / multi-line scan moves to utils::list_utils. MD030 and MD013 both call these. MD007 only governs marker placement, which reflow preserves rather than recomputes, so nothing is shared from it.

MD013 reads MD030's settings by loading MD030Config through the shared load_rule_config path (the same serde path MD030 uses) rather than re-parsing the TOML keys. Flagging for review: is this cross-rule config loading the right approach, or should the dependency be expressed some other way?

Assisted-by: Claude <noreply@anthropic.com>


rvben:
Thanks @chandlerc. I think the cross-rule config direction is sound; there's one edge case in the current `is_multi` decision.

Reading MD030 through `load_rule_config::<MD030Config>(config)` looks like the right approach. MD013 reflow needs the configured post-marker spacing so it doesn't fight MD030, and using the same serde loader plus `MD030Config::expected_spaces()` keeps the spacing policy in one place. Sharing `expected_spaces()` and the list-continuation scan is the right amount of extraction; I wouldn't add a separate list-policy abstraction just for these two rules.

The issue is in normalize mode when a plain multi-line list item is collapsed to one physical line. In the top-level list path, `is_multi` is:

```rust
let is_multi = is_multi_line_list_item(ctx, list_start + 1, lines)
    || /* predicted wrap */;
```

That first term is based on the source item. With:

```toml
[MD013]
reflow = true
reflow-mode = "normalize"
line-length = 200

[MD030]
ul-single = 1
ul-multi = 3
```

and:

```markdown
- first
  second
```

MD013 on its own rewrites the item as:

```markdown
-   first second
```

The item is now physically single-line, so MD030 expects `ul-single = 1` and rewrites it again to `- first second`. The full formatter converges (MD030 cleans up after MD013), so this isn't a true oscillation, but MD013 is producing a result that violates the MD030 config when run on its own or as an intermediate step.

The fix is to stop treating collapsible source continuation lines as proof the output should use multi-line spacing. The multi-line decision should follow the rewritten shape: true only if the emitted item still spans multiple lines, either because the prose wraps after reflow or because structural content remains (nested lists, code blocks, tables, admonitions, or multiple paragraphs). `ol-align-column` already ignores `is_multi` in `expected_spaces()`, so ordered lists keep working, and the MkDocs strict-indent branch is intentionally outside this path and should stay untouched.

A focused regression test is enough: non-default `ul-single = 1` / `ul-multi = 3`, normalize mode, a two-line plain bullet that collapses under a large line length, asserting the fixed output uses one space after the marker. Best as a direct rule-level test with the custom MD030 config, since the default idempotency proptests won't exercise this non-default spacing.

Happy to pair on the `is_multi` derivation if useful.

chandlerc:
Ooo, good catch. =D Fixed. Made one of the refactorings not make sense though -- no easy way to share the logic between MD030 and MD013 when we need to recompute during the reflow. PR is updated.

rvben (changes requested):
**Request changes** - the cross-rule design here is great, one correctness gap to close.

**Keep:** moving `expected_spaces` onto `MD030Config` as the shared spacing policy is the right call, and keeping MD013's rewritten-shape `is_multi` separate from MD030's source-shape one is well reasoned. With default MD030 the output is byte-identical, confirmed.

**Must fix - reflow silently flattens nested lists under non-default MD030.** When a wrapping parent item's marker widens (e.g. `MD030.ul-multi = 3` -> parent content column 4), a nested list/blockquote child keeps its original indent. `code_indent_shift` only reaches `Block::Code`/`Block::Table`; nested lists/blockquotes aren't `Block` variants here (they're reflowed by the outer loop independently), so they never get shifted. The child ends up under-indented relative to the parent's new content column and a CommonMark parser reparses it as a *sibling*, not a child.

Repro (`line-length = 44`, `MD030.ul-multi = 3`):

```markdown
- Outer bullet that is itself long enough to wrap across multiple lines for sure here.
  - Nested bullet that is also quite long and must wrap onto another line as well here.
```

After `--fix`:

```markdown
-   Outer bullet that is itself long enough
    to wrap across multiple lines for sure
    here.
  -   Nested bullet that is also quite long
      and must wrap onto another line as
      well here.
```

The original parses as outer -> nested; the output parses as two siblings (verified with a CommonMark parser). It's idempotent, so it won't oscillate - but it converges to a flattened structure, silently. The fix is to propagate the parent's indent-shift into the independently-reflowed nested list/blockquote subtree so the whole child moves with the parent.

**Must fix - test enshrines the detached state.** `test_reflow_nested_list_respects_md030_ul_multi` asserts the nested marker stays at `leading_spaces == 2` under a parent whose content column moved to 4. It should assert the child aligns to the parent's widened content column and that nesting is preserved (a parse/round-trip check would be ideal).

Scope: only non-default MD030 + nested lists/blockquotes; default config is unaffected. Everything else in the PR looks solid.

chandlerc:
Another good catch, should be fixed!

PR #694, opened 2026-06-26
fix(MD013): exempt standalone links with trailing punctuation and wrappers

chandlerc:
Standalone links (or images) that had trailing punctuation (e.g., `.`, `,`,
semicolons, colons, CJK punctuation) or were wrapped in emphasis, quotes,
or brackets/parentheses (e.g., `([text](url)),` or `**[text](url)**`) were
not recognized as standalone links. This caused them to be flagged as MD013
line length violations if they exceeded the limit, even though they are
structurally unsplittable.

Introduce a robust 5-step peeling loop in `is_link_with_optional_emphasis` that:
1. Checks if the string already matches a valid link/image pattern (match-first).
2. Peels matching wrappers from both ends if they are balanced (including
   parentheses `()` and brackets `[]` by verifying they have equal counts
   in the string).
3. Peels safe leading characters (non-structural wrappers like `(`, `{`,
   CJK brackets, quotes, emphasis).
4. Peels safe trailing characters (non-structural punctuation like `.`, `,`,
   `;`, CJK punctuation, closing quotes, emphasis).
5. Peels unsafe structural characters (`[`, `)`, `]`) only if they are
   unbalanced (extra) using a count-based heuristic (e.g., peeling leading `[`
   if `count('[') > count(']')`).

This safely handles both balanced and unbalanced wrappers (which occur when
parentheticals span multiple lines) for both forward and reverse patterns
without breaking the link's own structural delimiters.

Add comprehensive unit tests covering all these cases in `helpers.rs`
and a rule-level integration test in `tests.rs`.

### DIFF
--- a/src/rules/md013_line_length/helpers.rs
+++ b/src/rules/md013_line_length/helpers.rs
@@ -384,22 +384,7 @@ fn is_safe_leading(c: char) -> bool {
 fn is_safe_trailing(c: char) -> bool {
     matches!(
         c,
-        '.' | ','
-            | ';'
-            | ':'
-            | '!'
-            | '?'
-            | '}'
-            | '）'
-            | '｝'
-            | '】'
-            | '』'
-            | '」'
-            | '］'
-            | '、'
-            | '，'
-            | '；'
-            | '：'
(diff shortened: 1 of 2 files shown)


chandlerc:
Sorry for the force-push, the first version of this was really incomplete.

Updated to be a much more comprehensive fix to the overall problem.

Issue #271, opened 2026-01-06
Bug: `rumdl check --fix` is not idempotent, second run applies additional fixes


pygarap:
When I run `rumdl check --fix` twice on the same codebase (no manual changes between runs), the second run still applies additional fixes that were not done in the first run. The only differences are the changes made by the first `rumdl check --fix` run.

Expected: `rumdl check --fix` should be idempotent (running it again immediately should produce zero diffs).

Please investigate rule ordering or fix-application logic so a single `rumdl check --fix` run reaches a stable output.

rvben:
Thanks @pygarap! Fixed in v0.0.212 - the fix coordinator now re-checks all rules after each fix for idempotent results.

PR #695, opened 2026-06-27
fix(lsp): iterate formatting fixes to a fixpoint so one format pass converges

### LINKED ISSUE #145: formatting is not idempotent
Testcase:
```md
- Bullet 1
    - Sub-bullet with long text that is long enough that it is wrapped onto
      multiple lines.
- Bullet 2
```

After running the "Format document" command once, I get this:
```md
-   Bullet 1
      -   Sub-bullet with long text that is long enough that it is wrapped onto
          multiple lines.
-   Bullet 2
```

After running it a second time:
```md
-   Bullet 1
    -   Sub-bullet with long text that is long enough that it is wrapped onto
          multiple lines.
-   Bullet 2
```

And after a third time it reaches quiescence:
```md
-   Bullet 1
    -   Sub-bullet with long text that is long enough that it is wrapped onto
        multiple lines.
-   Bullet 2
```

This appears to be specific to the vscode extension. `rumdl check --fix` does not have this issue, presuambly due to https://github.com/rvben/rumdl/issues/271.

chandlerc:
Before, the LSP textDocument/formatting handler (VS Code's "Format Document") applied lint fixes in a single pass: it linted once and applied those fixes. When fixes cascade — e.g. MD030 widening a list marker, which then needs MD007 to re-indent the nested content and its continuation lines — one pass left work undone, so the editor had to run "Format Document" several times to converge. `rumdl check --fix` was unaffected because it iterates to a fixpoint through the FixCoordinator (rvben/rumdl#271).

After, the formatting handler routes through `apply_all_fixes`, the same iterative FixCoordinator engine the fix-all code action and the CLI use, so a single format pass reaches a fixpoint and the editor matches the CLI.

This is not a loss of LSP-specific behavior: `apply_all_fixes` performs the same config management the handler used to do inline — `merge_lsp_settings`, the VS Code enable/disable rule overrides (`apply_lsp_config_overrides`), and per-file flavor — and additionally honors file excludes and per-file-ignores, which the single-pass path skipped. Only the inline duplicate of that logic is removed. Range formatting, which delegates here, is covered too.

Add LSP regression tests: one formatting pass is idempotent and equals the `rumdl check --fix` output (the reported case and nested lists with continuation at both levels), and the formatter still honors an LSP disable-rules override.

Resolves rvben/rumdl-vscode#145

Assisted-by: Claude <noreply@anthropic.com>

### DIFF
--- a/src/lsp/server.rs
+++ b/src/lsp/server.rs
@@ -17,8 +17,6 @@ use crate::config::{Config, is_valid_rule_name};
 use crate::discovery::{ExcludeMatchers, is_markdown_extension};
 use crate::lsp::index_worker::IndexWorker;
 use crate::lsp::types::{IndexState, IndexUpdate, LspRuleSettings, RumdlLspConfig};
-use crate::rule::FixCapability;
-use crate::rules;
 use crate::workspace_index::WorkspaceIndex;
 
 /// Maximum number of rules in enable/disable lists (DoS protection)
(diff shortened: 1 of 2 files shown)


chandlerc:
BTW, huge thanks for all the response merges! And sorry for the deluge of fixes as we are adopting the tool, but _really_ happy this tool exists and is so easy to contribute back to.

rvben:
Thanks @chandlerc for the PRs! They help in making rumdl better for everyone and that's great 💪 really appreciated 🙂

PR #706, opened 2026-07-02
docs(nvim): Native Neovim LSP integration & configuration examples

kikefdezl:
Hello!

I struggled with the current docs to configure global settings in `rumdl` as LSP server in Neovim. I had to go into the internals to figure out the contract. 

These examples should make it much easier to understand how global configuration options can be passed. Specifically what I was missing is the following table in the `vim.lsp` setup:

```lua
settings = {
    rumdl = {
      lineLength = 100,
    }
  }
```

### DIFF
--- a/docs/lsp.md
+++ b/docs/lsp.md
@@ -85,6 +85,11 @@ vim.lsp.config("rumdl", {
   cmd = { "rumdl", "server" },
   filetypes = { "markdown" },
   root_markers = { ".git" },
+  settings = {
+    rumdl = {
+      lineLength = 100,
+    },
+  },
 })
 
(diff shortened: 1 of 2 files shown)


rvben:
Thanks @kikefdezl! This documents a contract that wasn't user-facing before. Confirmed it works end to end: the server parses the `rumdl` section from `didChangeConfiguration`, so `settings = { rumdl = { lineLength = 100 } }` maps onto the global line length. Merging - I'll do a small follow-up to align this with the existing `init_options` settings reference.

kikefdezl:
Great! Thank you for your work!

PR #713, opened 2026-07-11
test(action): restore malformed fixture and widen test-action.yml paths filter

soy-chrislo:
Follow-up from the discussion on #712.

`__tests__/fixtures/malformed/bad.md` stopped being malformed as a side effect of 497ab922 back in April, an unrelated commit whose message says "bad.md fixture corrected to valid markdown", but that file exists specifically to be malformed for these tests. This restores it.

Nothing caught the breakage for months because `test-action.yml` only triggered on changes to `action.yml`/`scripts/rumdl-action.sh`. Added `__tests__/fixtures/**` and the workflow file itself to the `paths` filters so this class of breakage triggers CI going forward.

**What's tested:**
- Opened a disposable PR against my own fork's main to confirm the widened paths filter actually fires on `pull_request` (it does)
- All 13 `Test Action` jobs pass with both changes combined, including the 4 that were failing on unmodified `main` (`test-args`, `test-expected-failure`, `test-combined-informational-with-output`, `test-output-file-with-violations`)

### DIFF
--- a/.github/workflows/test-action.yml
+++ b/.github/workflows/test-action.yml
@@ -2,9 +2,9 @@ name: Test Action
 on:
   push:
     branches: [main]
-    paths: ["action.yml", "scripts/rumdl-action.sh"]
+    paths: ["action.yml", "scripts/rumdl-action.sh", "__tests__/fixtures/**", ".github/workflows/test-action.yml"]
   pull_request:
-    paths: ["action.yml", "scripts/rumdl-action.sh"]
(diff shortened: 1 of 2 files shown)


rvben (approved):
Verified end to end: the restored fixture trips MD018, MD041, and MD047, and still fails with MD018 disabled, which is what test-args needs. The upstream Test Action run on this PR is all green, and the fact that it triggered at all confirms the widened paths filter fires. The original pre-April fixture had a `# Bad markdown` title line, so this isn't byte-identical, but it's functionally equivalent for every job that uses it. Merging this first so #712 can go green.

PR #714, opened 2026-07-11
perf(reflow): optimize inline code mask calculation using pre-extracted spans

chandlerc:
Pre-extract code span ranges using pulldown-cmark instead of manually looping through backticks in compute_inline_code_mask.

This resolves a potential quadratic O(N^2) complexity bottleneck when computing the inline code mask on lines with many unclosed backticks, achieving guaranteed O(N) linear time complexity.

### DIFF
--- a/src/utils/text_reflow.rs
+++ b/src/utils/text_reflow.rs
@@ -123,61 +123,31 @@ pub fn normalize_reference_label(label: &str) -> String {
 /// Build a boolean mask indicating which character positions are inside inline code spans.
 /// Handles single, double, and triple backtick delimiters.
 fn compute_inline_code_mask(text: &str) -> Vec<bool> {
+    let code_spans = extract_code_spans(text);
     let chars: Vec<char> = text.chars().collect();
-    let len = chars.len();
-    let mut mask = vec![false; len];
-    let mut i = 0;
-
-    while i < len {
-        if chars[i] == '`' {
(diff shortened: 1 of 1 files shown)


rvben (approved):
I reviewed this differentially: built main and this branch, ran sentence-per-line reflow over 90+ real docs plus crafted edge cases (nested and unequal backtick runs, multi-line spans, code in link titles); outputs are byte-identical and a second pass is idempotent. On an adversarial 200KB paragraph of strictly increasing backtick runs this is ~8x faster in a dev build (3.05s to 0.39s), so the superlinear behavior was real and is gone.

Two behavior changes exist, both matching CommonMark where the old scanner didn't: escaped backticks no longer mask sentence splits (the spec says they're literal, so the old masking was wrong), and backticks inside HTML comments lose what was only accidental protection (comments without backticks were never protected). Neither triggers on anything in docs/.

Non-blocking thought: this file now runs three pulldown passes per text (code spans, emphasis, links); consolidating into one offset-iter pass extracting all three span kinds would be a natural follow-up. Nice fix.

chandlerc:
> Non-blocking thought: this file now runs three pulldown passes per text (code spans, emphasis, links); consolidating into one offset-iter pass extracting all three span kinds would be a natural follow-up. Nice fix.

I have a few more fixes queued up here, but will add consolidating it into one pass after they land.

PR #724, opened 2026-07-13
fix(reflow): support multiple backticks and optimize code span parsing

chandlerc:
Update parse_markdown_elements_inner to identify and extract inline code spans using pre-extracted ranges from pulldown-cmark instead of manual character-by-character single-backtick matching.

This fixes the parser to correctly handle code spans delimited by multiple backticks (e.g., double backticks) as atomic code elements conforming to the CommonMark spec. It also eliminates the custom backtick parser loop, ensuring code spans are parsed in linear O(N) time. Stored Element::Code representation is updated to retain its original delimiters, and its Display implementation is updated to render it raw.

### DIFF
--- a/src/utils/text_reflow.rs
+++ b/src/utils/text_reflow.rs
@@ -645,7 +645,7 @@ impl std::fmt::Display for Element {
             Element::HugoShortcode(s) => write!(f, "{s}"),
             Element::AttrList(s) => write!(f, "{s}"),
             Element::MystRole(s) => write!(f, "{s}"),
-            Element::Code(s) => write!(f, "`{s}`"),
+            Element::Code(s) => write!(f, "{s}"),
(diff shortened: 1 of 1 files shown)


rvben:
Verified and landed on main as 81944c5a, cherry-picked with your authorship preserved (the branch had conflicted with the #728 reflow fix that touched the same test module, so I resolved locally rather than round-tripping a rebase).

Verification notes: on main, ``double backtick`` spans got split mid-span during reflow and everything after an unclosed backtick was swallowed as unwrappable text; with your change both stay atomic, and a link following a stray backtick wraps correctly. Full dev suite passed locally (11k+ tests).

The finder's rescan-from-start over the code spans was replaced with a monotonic cursor index in a30a4f97, together with the same pattern for the link and emphasis scans. Thanks for the fix!

PR #725, opened 2026-07-13
perf(reflow): implement cached match lookups to prevent quadratic suffix scanning

chandlerc:
Introduce absolute-offset based caching for regex and character search operations (wiki links, display math, inline math, emoji, HTML entity, Hugo shortcode, HTML tags, curly brackets) in parse_markdown_elements_inner.

This resolves worst-case quadratic O(N^2) complexity on long lines by ensuring we only re-run regex matching when the parsing cursor advances beyond the cached match start offset, rather than re-running regexes on the remaining suffixes on every iteration.

### DIFF
--- a/src/utils/text_reflow.rs
+++ b/src/utils/text_reflow.rs
@@ -1044,6 +1044,12 @@ fn myst_role_len_at(text: &str) -> Option<usize> {
 /// 5. Reference links [text][ref] - before shortcut references
 /// 6. Shortcut reference links [ref] - detected last to avoid false positives
 /// 7. Other elements (code, bold, italic, MyST roles, etc.) - processed normally
+#[derive(Clone, Copy, Debug)]
+struct PatternMatch {
+    start: usize,
+    end: usize,
+}
+
 fn parse_markdown_elements_inner(
     text: &str,
     attr_lists: bool,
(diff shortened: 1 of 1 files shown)


rvben:
Landed on main as 1cc8d2bc, cherry-picked with your authorship preserved (the branch conflicted with both #724 and the #728 reflow fix, so I resolved locally).

The perf win is dramatic and real: I measured the pathological input (80k chars of backtick runs) through full `reflow_line` at 374 seconds on main vs 0.42 seconds with your caching - that was a genuine hang vector for the LSP on pasted minified content. The caching design is sound for all the plain regexes, since without lookaround a match's existence at a position is independent of search start.

One regression was fixed forward in a30a4f97: `INLINE_MATH_REGEX` starts with the lookbehind `(?<!\$)`, which IS slice-start-sensitive - the old per-iteration search accepted a `$` at the cursor whose real predecessor was a `$` (the lookbehind sees nothing before the slice start), while the cache replayed an earlier-anchored search that rejected it. Concretely, `$$a$$$bb cc dd$ x` at width 8 lost the inline span's atomicity and wrapped it apart. The fix re-searches whenever the cursor sits directly after a `$`, which is exactly the divergence window, so the speedup is preserved. Also moved `PatternMatch` above the parser's doc comment, which had ended up attached to the struct.

Thanks for the thorough work on this!

PR #732, opened 2026-07-13
perf(reflow): merge emphasis and code span extraction into a single cmark pass

chandlerc:
Update parse_markdown_elements_inner to extract both emphasis spans and
code spans in a single shared pulldown-cmark parser run.

This reduces the number of separate markdown parsing passes per paragraph from
three to two, avoiding redundant tokenization overhead. The link extraction pass
must remain separate due to broken-link resolution shifting delimiter pairings in
corner cases. A standalone extract_code_spans is retained only for the mask
calculation utility.

Note that this is based on #731 and #730 -- only the last commit is for review here.

### DIFF
--- a/src/utils/text_reflow.rs
+++ b/src/utils/text_reflow.rs
@@ -746,16 +746,21 @@ struct EmphasisSpan {
 /// - GFM strikethrough (~~text~~)
 ///
 /// Returns spans sorted by start position.
-fn extract_emphasis_spans(text: &str) -> Vec<EmphasisSpan> {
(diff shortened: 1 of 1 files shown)


chandlerc:
BTW, let me know if you'd rather I handle a stack of changes in some other way -- using the last commit has worked for me in some other projects, but happy to adopt whatever workflow works for you.

rvben:
This workflow worked well on my side: I reviewed the last commit of each PR as suggested and squash-merged the three in order. Feel free to keep stacking this way.

PR #736, opened 2026-07-15
feat(reflow): support breaking within emphasis spans

chandlerc:
Adds a new configuration option `emphasis_spans` to MD013 (Line Length) config (which translates into `emphasis_spans` in `ReflowOptions`), matching the naming of `code_spans`.

When enabled (`true`), emphasis spans (bold, italic, strikethrough) are no longer treated as atomic units. Instead, they are split word-by-word like normal text and wrapped across line boundaries if they don't fit on the current line.

By default, this option is `false` to preserve the original behavior of keeping short emphasis spans atomic.

Also includes:
- Refactoring of internal `cascade_split_line` signature to take `&ReflowOptions` directly, simplifying option propagation.
- Unit tests verifying the wrapping of emphasis spans with different markers.
- Documentation for the new `emphasis-spans` configuration option in `docs/md013.md`.

### DIFF
--- a/docs/md013.md
+++ b/docs/md013.md
@@ -52,6 +52,7 @@ and work with in various contexts.
 line-length = 100  # Maximum characters per line (default: 80)
 code-blocks = false  # Don't check code blocks (default: true)
 code-spans = false  # Don't flag lines whose only overflow is an unbreakable inline code span (default: true)
+emphasis-spans = false  # Allow reflow breaking inside emphasis/strong/strikethrough spans (default: false)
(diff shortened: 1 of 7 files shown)


chandlerc:
This... was not actually the thing I wanted. Please disregard, working on a fresh PR that it _actually_ what I'm looking to achieve here.

chandlerc:
Just to be clear -- this wasn't the feature I actually needed. I'm working on a replacement that is. Do you still want this? Mostly so I know whether to rebase the feature I actually want on top or not.

PR #737, opened 2026-07-15
fix(tests): resolve String addition compilation errors under Rust 1.96

chandlerc:
Fix compilation failures under Rust 1.96.0+ caused by attempting to add `&String` to `String` (e.g. `cannot add &String to String`).

Resolved by converting string repetitions to borrowed string slices using `.as_str()` before addition. This ensures the entire integration test registry compiles successfully.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -92,7 +92,7 @@ requires = ["setuptools>=42", "wheel"]
 build-backend = "setuptools.build_meta"
 
 "#;
-        let content = basic_content.to_owned() + &config_content;
+        let content = basic_content.to_owned() + config_content.as_str();
 
         match fs::write("pyproject.toml", content) {
             Ok(()) => {
--- a/src/rules/md013_line_length.rs
+++ b/src/rules/md013_line_length.rs
@@ -2657,7 +2657,7 @@ impl MD013LineLength {
                                         // First line of first block gets marker
                                         result.push(format!(
                                             "{marker}{}",
-                                            " ".repeat(orig_indent - marker_len) + content
+                                            " ".repeat(orig_indent - marker_len) + content.as_str()
                                         ));
                                         is_first_block = false;
                                     } else if content.is_empty() {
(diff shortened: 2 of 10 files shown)


rvben:
Thanks for the PR. I tried to reproduce this and could not: `String + &String` compiles on stable rustc 1.96.0 (the toolchain this repo pins via rust-toolchain.toml), on the current nightly 1.99.0 (2026-07-14), and the full test suite including the integration registry builds green on 1.96.0 in CI and locally. The `impl Add<&str> for String` plus deref coercion path this idiom relies on is unchanged upstream.

Could you share your exact `rustc --version` and the full error output? If your toolchain carries local patches this may be specific to it. I'd rather not take the churn without a rationale that reproduces on upstream Rust, but happy to revisit with a reproduction.

chandlerc:
Doh, sorry about this. I had just updated `rustc` and so this was a plausible reason, but it wasn't accurate.

The real issue is that in a subsequent change I'm looking at adding use of a Rust crate that transitively pulls in the `smartstring` crate. That crate in turn includes another `Add` impl that breaks these `+`s.

I can fold the fix (this commit or an alternative) into the PR that adds the thing that uses them so it makes more sense.

A question: any preference about how we do this? There is the approach in this PR, but we could also more aggressively use `format!` to replace all the `+`s. Which approach would you prefer?

rvben:
Thanks for the follow-up, that explains it. Folding it into the dependency PR sounds right, since there's nothing to merge standalone (upstream Rust compiles the current code fine).

On style: I'd keep the minimal `.as_str()` form you already have, limited to the sites the new `Add` impl actually breaks. `format!` reads fine in tests, but the non-test sites (md013, anchor_styles) benefit from `+` reusing the left-hand buffer, and I'd rather avoid a broader style migration motivated by a transitive dependency. Out of curiosity, which crate is coming in that pulls smartstring? That context will help when reviewing the follow-up PR.

PR #758, opened 2026-07-24
feat(invisiblechars): new rule MD084 for detecting invisible characters

guillp:
# What

I propose this relatively simple new rule that reports on invisible characters such as Zero-Width spaces, No-Break spaces, Soft-Hyphens, and other Unicode characters that are technically invisible because they have no immediate representation on screen.

# Why

Those characters are rarely used intentionally but are rather typically copy/pasted by mistake from Web sources.
They can cause rendering bugs, lead to confusing IDE behaviors, and even block `rumdl` from applying auto-fixes (e.g. for MD060 table formatting).

# How

To avoid triggering too many false-positives, the rule has 2 modes:

- default: triggers only on:
    - multiple consecutive invisible characters
    - any invisible character at the start or end of a line
    - invisible characters adjacent to any visible whitespace character
- strict: it will trigger on any invisible character

Both modes have an allow list to allow specific characters for real intentional uses of those special chars.

The available Auto-fix is to remove the detected invisible chars.

---

Let me know what you think about this :)


guillp:
(sorry for the force pushes, my email was misconfigured)

rvben (changes requested):
Thanks for this - the concept is valuable and the implementation is clean and well-tested. Detecting invisible/zero-width characters (including the bidi controls that enable Trojan-Source style attacks) is a genuinely useful, differentiated rule, and I like the opt-in default plus the allow-list.

One correctness issue blocks merge, plus a docs fix.

## Must fix: TAB (U+0009) is treated as invisible

`is_invisible_char` matches `0x0000..=0x001F` (`src/rules/md084_invisible_characters.rs`), which includes TAB. LF/CR never reach line content (stripped by `content.lines()`), but TAB does, so ordinary tab indentation gets flagged and the autofix deletes it:

```
$ printf 'text\n\tindented\n' | rumdl check --enable MD084
:2:1: [MD084] Invisible character U+0009 detected at line boundary [*]

# rumdl fmt --enable MD084 then rewrites the file:
#   before: "\n\tindented"  (0a 09 69 ...)
#   after:  "\nindented"     (0a 69 ...)   <- tab silently removed
```

That corrupts tab-indented content (indented code blocks, list continuations), and in strict mode every tab in the file is flagged. It also collides with MD010 (no-hard-tabs), which fires on the same position with the opposite fix (tab -> spaces vs delete).

Suggested fix: exclude TAB from the range, e.g.

```rust
0x0000..=0x0008
| 0x000A..=0x001F   // C0 controls, minus TAB (U+0009)
```

(LF/CR are already gone by the time lines are scanned, so only TAB needs excluding; tabs are MD010's domain.) A small regression test asserting a leading tab is not flagged would lock this down - none of the current tests exercise a tab.

## Should fix: docs contradict the code on strict-mode autofix

`docs/md084.md` Notes says "Auto-fix is provided only in default mode... In strict mode, findings are report-only," but strict mode does provide an autofix now (commit "strict mode has auto-fix too", and `test_strict_flags_any_invisible_character` asserts it). Please update the Notes section to match.

## Nits

- `docs/md084.md`: "block rumdl for applying auto-fixes" -> "from applying"; "invsible" -> "invisible".
- In strict mode, VS16 (U+FE0F) inside emoji like ❤️ would be flagged; probably worth a line pointing users at `allow` for that case.

Everything else looks good to me: robust allow-list parsing, correct multi-byte column reporting, and no off-by-one in the run/boundary/adjacency logic. Happy to approve once the TAB handling is sorted.

guillp:
Thanks to you @rvben . I fixed most of your findings, excepted for the VS16 inside emojis. It is probably worth a specific logic to allow emojis even in strict mode.

rvben:
Thanks @guillp - the TAB fix is confirmed. I verified it behaviorally on your branch rather than just reading the diff:

- `\tindented` is clean in both default and strict mode
- `rumdl fmt --enable MD084` leaves the `09` byte intact
- Positive controls (U+200B at a line boundary, U+202E adjacent to whitespace) still fire correctly

Merging this.

On VS16 you were right that it needs dedicated logic, and it turns out to be broader than the strict-mode nit I filed. It fires in **default** mode too, where the autofix is lossy:

```
$ rumdl check --enable MD084 warn.md
warn.md:1:4: [MD084] Invisible character U+FE0F detected adjacent to visible whitespace [*]

# rumdl fmt --enable MD084
before: > ⚠️ Note: this is important
after:  > ⚠ Note: this is important
```

Any emoji whose VS16 lands at a line boundary or next to a space hits this: `⚠️`, `❤️`, `✔️`, `▶️`, or a heading ending in an emoji. Stripping VS16 flips the character from emoji to text presentation, so it is a visible glyph change rather than a no-op. My review scoped that as a strict-mode nit, which undersold it.

I'll take it as a follow-up on main before MD084 goes out in a release, so the rule ships correct on its first appearance. Nothing further needed from you here.

Thanks for the contribution - detecting bidi controls and zero-width characters is a genuinely useful addition to the rule set.

PR #763, opened 2026-07-27
feat(md088): normalize quotes and dashes to ASCII

guillp:
## What

- a simple opt-in rule that replaces fancy Unicode quotes, and optionally dashes, by their ASCII equivalents.
- those characters are more typographically correct but they offer very little readability gain and are a source of variability and noisy diffs.

## Why

Normalizing quotes and dashes to `'`, `"` and `-` keeps Markdown content predictable and avoids hidden character drift across contributors and platforms.

This is one of the fixes applied by [`ftfy`](https://github.com/rspeer/python-ftfy/) / [`plsfix`](https://github.com/kevinhu/plsfix).

## How

The rule is opt-in and only checks for Unicode quotes and dashes chars in prose. Doesn't touch inline, fenced or indented code blocks.
The autofix is a simple character replacement by the ASCII equivalent `"`, `'` or `-`.
An allow list is configurable to ignore some intentionally used characters.


guillp:
@rvben : Please let me know what you think about this, but don't merge yet. I intend to generalize the rule to cover more Unicode characters that look like their ASCII equivalents, like —/- , ‹/< , etc. . 
This new rule will avoid generating more Mojibake since it avoids commonly used Unicode characters.

rvben (commented):
Thanks for the rework, this is much closer. The MD086 renumber, the opt-in default, the `allow` list, and suppressing the fix inside links and bare URLs all land well. Promoting `is_in_bare_url` onto `LintContext` is a good cleanup that MD044 gets to share.

I built `5f17c04` and ran everything below through the CLI against that build. Four things block merging, and then there is a design question about the character set that I think is the real decision here.

## 1. `RuleCategory::Emphasis` makes the rule invisible in the real lint path

`should_skip_rule` (`src/lib.rs:203`) drops every `Emphasis` rule unless some line contains `*` or `_` (`src/lib.rs:184`). MD086 therefore does nothing on the text it most cares about:

```console
$ printf 'It’s a “quote”.\n' > a.md
$ rumdl check --no-config --no-cache --enable MD086 a.md
Success: No issues found in 1 file (3ms)

$ printf 'It’s a “quote” with *emphasis*.\n' > b.md
$ rumdl check --no-config --no-cache --enable MD086 b.md
b.md:1:3: [MD086] Look-alike character ’ (U+2019) should be replaced with ' [*]
b.md:1:8: [MD086] Look-alike character “ (U+201C) should be replaced with " [*]
b.md:1:14: [MD086] Look-alike character ” (U+201D) should be replaced with " [*]
```

Same text, one asterisk apart. The `rules.json` entry in this PR already says `category: whitespace`, which is the right answer: `Whitespace` is in the always-run set, and MD084 (`invisible-characters`), the nearest analogue, uses it for exactly this reason.

The unit tests miss this because they call `rule.check(&ctx)` directly and bypass the driver, so it is worth adding one test that goes through the normal lint entry point. That is the only level where this is observable. (Every repro below carries a stray `*x*` purely to work around it.)

## 2. U+2039 / U+203A become `<` and `>`, which turns text into markup

```console
$ printf 'French quote: ‹b›bold‹/b› here. *x*\n' > guil.md
$ rumdl fmt --no-config --no-cache --enable MD086 guil.md
$ cat guil.md
French quote: <b>bold</b> here. *x*
```

The rendered output changes, not just the source (markdown-it, commonmark preset):

```text
before: <p>French quote: ‹b›bold‹/b› here.</p>
after:  <p>French quote: <b>bold</b> here.</p>
```

U+2039/U+203A are single guillemets, the French quotation marks. They are not look-alikes for ASCII angle brackets in any useful sense, and this is the one mapping in the rule that can inject HTML into a document.

## 3. The homoglyph table strips ordinary diacritics

With `enable-latin-homoglyphs = true`:

```text
before: café Müller Łódź żółw Việt
after:  cafe Muller Łodź zołw Việt
```

`Łódź` is the clearest tell: `ó` is rewritten while `Ł` and `ź` in the same word survive. That is not a coherent class, and the table shows why. Of its 54 entries, only 22 are genuine cross-script confusables (14 Cyrillic, 4 Greek, 4 Armenian). The other 32 are ordinary Latin letters with diacritics:

```text
ạ ą ä à á ƈ ċ ɗ ẹ ė é è ġ í ì ï ʝ ḷ ȯ ọ ỏ ơ ó ò ö ʂ ü ú ù ý ʐ ż
```

The `é` in `café` is not a look-alike for `e`, it is the correct spelling. Rewriting it corrupts French, German, Polish, Vietnamese and a long tail of other languages, silently, under `rumdl fmt`.

## 4. Math and typography get rewritten

```text
before: 3 × 4 ÷ 2, f′(x), 5′10″.
after:  3 x 4 / 2, f'(x), 5'10".
```

`×` and `÷` are operators and `′`/`″` are prime marks (derivative, feet and inches, minutes and seconds). None of them are accidental substitutions for `x`, `/`, `'` or `"`; they are the intended characters, and inside `$...$` the rewrite changes what the expression means.

## 5. U+02BC is a letter

U+02BC (modifier letter apostrophe) is still in the default punctuation set, so it is rewritten with the rule merely enabled:

```text
before: Ukrainian сʼято and Hawaiʻi.
after:  Ukrainian с'ято and Hawaiʻi.
```

In Ukrainian, U+02BC is a letter, not punctuation. The rule already leaves U+02BB (the Hawaiian okina) alone, so it implicitly agrees that modifier letters are not quotes; U+02BC just needs to follow the same logic.

## The design question

All five findings are instances of one thing, so I would rather settle the principle than patch five entries:

**Only replace characters that carry no meaning of their own.** A character that is a letter in some orthography, or an operator in some notation, is never a look-alike. It is somebody's correct text.

By that line:

| group | codepoints | verdict |
|---|---|---|
| smart quotes | `‘ ’ ‚ ‛ “ ” „ ‟` | keep, this is the rule's core |
| dashes | `‒ – — ― −` | opt-in at most, these are deliberate typography rather than broken hyphens |
| guillemets | `‹ ›` | drop, injects markup |
| operators, primes | `× ÷ ′ ″` | drop, changes meaning |
| modifier letters | `ʼ` U+02BC | drop, it is a letter |
| cross-script confusables | the 22 Cyrillic/Greek/Armenian entries | keep, behind the existing option |
| Latin diacritics | the 32 above | drop |

That leaves a rule I would be comfortable enabling by default one day: quotes normalized, with confusable detection there for people who want it.

One further thought on the confusables half, since its purpose is genuinely different from the rest. A Cyrillic `а` in Latin text is either a paste accident or a deliberate spoof, and in the spoof case an autofix quietly destroys the evidence rather than surfacing it. Reporting those without a fix, or splitting them into their own rule, may serve users better than inheriting `FullyFixable` from the quote half. I am happy either way, but it is worth deciding deliberately.

Nit: `alias = "enable_latin_homoglyphs"` is redundant. `normalize_key` kebab-cases every key before serde sees it, so the snake_case spelling already resolves to the canonical name. I checked that it does work, so this is dead weight rather than a bug.

guillp:
My initial intention when implementing the homoglyph replacement was to easily spot confusable characters, which are almost certainly mistakes, so are worth being linted. And since most documentations I work with are in English (with a simple latin charset), I implemented this simple list of latin lookalikes. But I forgot that a lot of Latin languages have accentuated or other derived letters, so the rule does not work well.

For this homoglyph rule to work better, `rumdl` should actually be aware about which language the documentation is written in, so it knows which letters are supposed to be there and which should not. I am not sure if you want `rumdl` to do that. And a simple spellchecker should already catch most of those issues. So this is probably not worth the complexity.

On the other hand, having a simple and fast tool (such as rumdl) that checks URLs in untrusted markdown files and spots potential confusables could be useful, since those URLs are very likely to be malicious. So this (different) rule would be a sort of security control. What do you think about this?

I'll refocus this rule on quotes and optionally dashes.

rvben:
Thanks for the PR @guillp, and for coming back with the honest reassessment - that saved us both a round.

Your own read is the right one: drop the Latin-lookalike half. The issue isn't an incomplete list, it's that "this letter doesn't belong here" is a claim about the document's language. French, Spanish, Turkish and Vietnamese prose is full of correct non-ASCII letters, so the rule flags valid content and, with an autofix, rewrites it. No amount of list tuning fixes that.

Quotes and optionally dashes is a good rule: a closed, language-independent set with an unambiguous mapping. Keep it opt-in with your allow list, and make dashes separately configurable from quotes - `-` for `—` is usually fine, `-` for `–` in a numeric range is a meaning change.

The URL idea is worth doing, and it's better scoped than prose homoglyphs for the same reason: a host is a constrained grammar, not natural language, so non-ASCII there is intrinsically suspicious, and IDNA plus TR#39 confusables give you real prior art instead of a hand-written list. Two conditions: report only, never autofix (rewriting a link destination is where a wrong guess sends the reader somewhere else), and its own rule number rather than a mode of the quotes rule.

One blocker before your next push: **MD086 is taken** (Comments should be closed), and so is MD087. Use **MD088** - renaming later breaks anyone's config.

Nice shape overall: with MD083 (mojibake) and MD084 (invisible characters), that's a coherent family - bytes that got mangled, characters you can't see, and characters that render as something other than what they are.

guillp:
@rvben : rebased on current main, changed to MD088, simplified to handle only quotes and dashes.
Also moved a few unicode specific functions from MD084 to a dedicated `utils::unicode` namespace, so I could reuse them in this new rule and applied matching changes in MD084.

guillp:
@rvben  Thanks for merging and fixing the leftover issues that I had no time to fix myself 👍

rvben:
Thanks again for your work @guillp, really appreciated! 🚀

rvben:
Merged, and released in [v0.2.51](https://github.com/rvben/rumdl/releases/tag/v0.2.51). Thank you for the rule and for the test coverage that came with it - the design was right: opt-in, ASCII-only output, and code spans and code blocks left alone.

Three follow-ups landed on top before the tag, so MD088 ships correct on its first appearance:

**`56dbf0fe` - modifier letters and math notation are left alone.** U+02BC MODIFIER LETTER APOSTROPHE is a `Modifier_Letter`, not punctuation. Crimean Tatar, Nenets and Chechen spell words with it, so normalizing rewrote `Qaradenʼiz` to `Qaraden'iz` and changed the word rather than the typography. U+02BB (the Hawaiian okina) was already absent from the table, which is what surfaced the inconsistency - both are letters, so both stay. Characters inside a math context are skipped for the same reason: a prime there is notation, not a quotation mark.

**`557c490b` - the docs page was headed MD086**, and both config examples used `[MD086]`. Anyone copying one got three `Unknown option for rule MD086` warnings and MD088 never ran. The alias was written `normalize-quotes-dashes` where the registry accepts `quotes-dashes`, and the double prime was listed as U+2032 rather than U+2033.

**`62600652` - registered in the frozen rule sets.** The opt-in rule set, the rule total and the configurable-rule count are deliberate stability surfaces with tests asserting them, so a new opt-in rule has to be added there or the next unrelated change trips them. `docs/rules.md` listed MD088 only in the per-category table.

PR #764, opened 2026-07-27
feat(invisiblechars): extend MD084 with flagging deprecated unicode chars

guillp:
## What

Extends MD084 with also flagging deprecated Unicode characters `0x0340`, `0x0341` and `0xFFFC`. Those are reported but are not autofixable since they are visible.

I used http://www.unicode.org/reports/tr20/tr20-9.html#Suitable as reference. The other codepoints mentioned there are invisible characters so are already handled by the initial rule.

(+ a few refactoring to the test cases)

### DIFF
--- a/docs/md084.md
+++ b/docs/md084.md
@@ -1,4 +1,4 @@
-# MD084 - Invisible Unicode characters
+# MD084 - Invisible and deprecated Unicode characters
 
 Aliases: `invisible-characters`
 
@@ -17,6 +17,7 @@ By default, it reports only suspicious patterns:
 - multiple consecutive invisible characters
 - any invisible character at the start or end of a line
 - invisible characters adjacent to any whitespace character
+- deprecated visible Unicode characters (not auto-fixable)
 
 It doesn't report invisible character surrounded on both side by visible characters,
 since that is how those invisible characters are typically used.
@@ -67,6 +68,14 @@ The list of non-visible characters that this rule will trigger on is:
 | 0x1BCA0..=0x1BCA3 | Shorthand format controls                                  |
 | 0x1D173..=0x1D17A | Musical symbol format controls                             |
 | 0xE0000..=0xE0FFF | Tags block + Variation Selectors Supplement                |
+
+The list of deprecated Unicode characters:
+
+| Unicode Codepoint | Description                  |
+| ----------------- | ---------------------------- |
+| 0x0340            | COMBINING GRAVE TONE MARK    |
+| 0x0341            | COMBINING ACUTE TONE MARK    |
+| 0xFFFC            | OBJECT REPLACEMENT CHARACTER |
 
 ## Why this matters
 
--- a/docs/md084.md
+++ b/docs/md084.md
@@ -7,20 +7,23 @@ Aliases: `invisible-characters`
 
 ## What this rule does
 
-This rule detects invisible Unicode characters that can cause copy/paste and rendering bugs,
+This rule detects:
+
+- invisible Unicode characters that can cause copy/paste and rendering bugs,
 confuse readers, or block `rumdl` from applying auto-fixes.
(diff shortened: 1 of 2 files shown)


rvben (commented):
Good reference, and I want the U+FFFC half of this. Three notes, one of which is a free win and one of which is a gap in the existing rule that your reference points at.

**U+FFFC: yes, and reporting it without a fix is right.** It is a placeholder for an embedded object, so its presence means content was lost on the way into the file, and only the author knows what it was.

**U+0340 and U+0341 are fixable, safely.** They are canonically equivalent to U+0300 and U+0301, which every normalization form confirms:

```text
"a" + U+0340  ->  NFC U+00E0   NFD U+0061 U+0300   NFKC U+00E0   NFKD U+0061 U+0300
"a" + U+0341  ->  NFC U+00E1   NFD U+0061 U+0301   NFKC U+00E1   NFKD U+0061 U+0301
```

Rewriting U+0340 to U+0300 therefore cannot change the rendered text, which makes it a safer fix than most of what MD084 already does. It also means these two only ever show up in unnormalized text, since all four forms remove them.

**On the "deprecated" framing.** The current UCD `Deprecated` property does not cover any of the three codepoints in this PR. It is exactly U+0149, U+0673, U+0F77, U+0F79, U+17A3..U+17A4, U+206A..U+206F, U+2329, U+232A and U+E0001 (PropList.txt). TR20's comment column does say "Deprecated in Unicode" for U+0340..U+0341, but the criterion that unifies that table, and its title, is *characters not suitable for use with markup*. That name describes what the rule is doing and gives the next addition a principled home; "deprecated" sends a reader to the UCD property and misleads them.

**The gap.** If TR20 Table 3.1 is the criterion, one of its rows is already half-implemented, in the wrong half. MD084 has:

```rust
| 0xFFF0..=0xFFF8 // Interlinear annotation and reserved non-rendering specials
```

but the interlinear annotation characters are U+FFF9 (ANCHOR), U+FFFA (SEPARATOR) and U+FFFB (TERMINATOR). U+FFF0..U+FFF8 are unassigned. So the rule flags the reserved block and misses the characters its own comment names. On 0.2.46, with each file holding `text <the three codepoints> more`:

```console
$ rumdl check --no-config --no-cache --enable MD084 --config 'MD084.strict = true' fff9.md

Success: No issues found in 1 file (0ms)

$ rumdl check --no-config --no-cache --enable MD084 --config 'MD084.strict = true' fff0.md
fff0.md:1:6: [MD084] Invisible character U+FFF0 detected (strict mode) [*]
fff0.md:1:7: [MD084] Invisible character U+FFF4 detected (strict mode) [*]
fff0.md:1:8: [MD084] Invisible character U+FFF8 detected (strict mode) [*]

Issues: Found 3 issues in 1 file (0ms)
Run `rumdl fmt` to automatically fix 3 of the 3 issues
```

Extending that range to `0xFFF0..=0xFFFB` is a one-line change. Take it here if you want it, or leave it and I will do it separately.

Confirmed none of your three are flagged today, in either mode, so this is purely additive (`d3.md` holds `a`+U+0340, `c`+U+0341 and U+FFFC):

```console
$ rumdl check --no-config --no-cache --enable MD084 d3.md

Success: No issues found in 1 file (0ms)

$ rumdl check --no-config --no-cache --enable MD084 --config 'MD084.strict = true' d3.md

Success: No issues found in 1 file (0ms)
```

Happy to merge with U+FFFC unfixable, the tone marks fixable to their canonical equivalents, and the docs wording adjusted. It needs a rebase on `main` as well, it is currently behind.

guillp:
@rvben : Thanks for the feedback. All good points, all fixed:
- added replacements for U+0340 and 0+0341
- added all officially deprecated Unicode characters
- extended the invisible range `U+FFF0..=U+FFF8` to `U+FFF0..=U+FFFB`
- updated docs
- rebased on current `main`

Keep up the good work! 👍🏻

PR #766, opened 2026-07-27
feat(MD041): `allow_preamble` option for markdownlint parity

### LINKED ISSUE #761: MD041 `allow_preamble` option
### Problem or motivation

I tried to migrate from `markdownlint` to `rumdl` and the only blocker to make it a "drop in" replacement is the lack of an equivalent `allow_preamble` option for `MD041`.

### Proposed solution

Add support for an `allow_preamble` option in MD041, matching `markdownlint` behavior.

This would allow content such as front matter, comments, or other preamble text before the first top-level heading while keeping MD041 enabled.

### Alternatives considered

_No response_

### Feature category

Improve existing rule

### Example markdown

```markdown

```

### Contribution

### Additional context

_No response_

guillp:
# What

Add a `MD041.allow_preamble` option for feature parity with `markdownlint`. Fixes #761 .

With this option enabled, any prose before the first heading is skipped. So the first heading found must have the appropriate level. If there is no heading at all, it is still considered an issue.

### DIFF
--- a/docs/md041.md
+++ b/docs/md041.md
@@ -82,6 +82,7 @@ level = 1  # Heading level required (1-6, default: 1)
 front-matter-title = "title"  # Front matter field to use as title
 front-matter-title-pattern = "^(title|header):"  # Regex pattern to match title fields in front matter
 fix = false  # Enable auto-fix (default: false)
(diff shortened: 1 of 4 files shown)


guillp:
@rvben: you implemented it yourself, closing 👍🏻

PR #777, opened 2026-08-01
fix(md013): support links with nested formatting and unresolved reference links in exemptions

chandlerc:
Refines the MD013 (line length) exemption logic for standalone links to handle nested formatting (e.g. `[*text*](url)`) and unresolved reference links (e.g. `[text][]`).

This PR adds a regression test for nested brackets and code spans in link exemptions.

Depends on #781.

### DIFF
--- a/src/lint_context/link_parser.rs
+++ b/src/lint_context/link_parser.rs
@@ -141,9 +141,10 @@ pub(super) fn parse_links_images_pulldown<'a>(
         content,
         options,
         Some(|link: BrokenLink<'_>| {
+            let span_end = extend_collapsed_byte_end(content, link.link_type, link.span.end);
             broken_links.push(BrokenLinkInfo {
                 reference: link.reference.to_string(),
-                span: link.span.clone(),
+                span: link.span.start..span_end,
                 link_type: link.link_type,
             });
             None
--- a/src/rules/md013_line_length.rs
+++ b/src/rules/md013_line_length.rs
@@ -503,7 +503,7 @@ impl Rule for MD013LineLength {
                 // Lines whose only content is a link/image are exempt.
                 // After stripping list markers, blockquote markers, and emphasis,
                 // if only a link or image remains, there is no way to shorten it.
-                if is_standalone_link_or_image_line(line) {
+                if is_standalone_link_or_image_line(ctx, line_number) {
                     continue;
                 }
 
@@ -894,7 +894,7 @@ impl MD013LineLength {
             // A standalone link/image line is exempt from MD013 (non-strict mode),
             // so it must end the blockquote paragraph rather than be absorbed into
             // it, mirroring the top-level paragraph reflow boundary.
-            || (!strict && is_standalone_link_or_image_line(content))
+            || (!strict && is_standalone_link_or_image_line(ctx, line_num))
     }
 
     #[allow(clippy::too_many_arguments)]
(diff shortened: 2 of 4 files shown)


chandlerc:
Note: depends on #775

rvben:
Thanks for this @chandlerc, and for the three PRs alongside it - #775, #776 and #778 are all merged.

That means this one needs a rebase, and two files moved underneath it:

- `helpers.rs`: blockquote and list-marker stripping is now one shared `strip_structural_prefixes` that alternates until neither matches, since the two nest in either order.
- `link_parser.rs`: the link-text scan was quadratic (a per-character backwards walk over preceding backslashes) and now carries escape state forward. `tests/perf/rule_complexity_regression_test.rs` fails if it comes back, so extending the callback span has to keep the single forward pass.

I ran your cases against main. `- [link]` and `- > [link]` are already exempt, so those two tests are now regression guards rather than fixes (the blockquote one did fire on 0.2.47). Still reported at `line-length = 80`, `stern = true`, `ignore-link-urls = false`:

```markdown
-   [`[T; N]` builtin syntax](/proposals/p004682-....md#t-n-builtin-syntax)   -> MD013, 134 chars
-   [[T; N] builtin syntax](/proposals/p004682-....md#t-n-builtin-syntax)     -> MD013, 132 chars
```

So the nested-brackets and nested-code-span half is a real open bug, worth fixing.

Before I can merge:

1. Rebase, so the diff shows only what is still needed - some of the 580/197 lines is now overlap.
2. The new `proptest-regressions/.../block_builder.txt` seed means the block-builder change failed the idempotency proptest at least once. Checking it in is right, but tell me what the failure was.
3. Please split it. The list-reflow line-number fix and the binary-search optimization are independent of the exemption work, and separate PRs are far easier to review and to revert if one turns out wrong.
4. Say which reflow expectations fail on current main. Those tests assert exact output, and I want to land them knowing which are new behavior.

chandlerc:
Thanks for the review and the detailed feedback!

I think I have addressed all the points and restructured the stack as requested. Here is a summary of the updates:

1. PR stack split:

   To make review and potential reverts easier, I have split the large change into a stacked sequence of 4 PRs. Each PR adds exactly one commit on top of the prior:
   - **PR #780** (New): This addresses the list-reflow line-number fix.
   - **PR #781** (New): This replaces the regex scans with O(log N) binary searches on parsed elements.
   - **PR #777** (This PR, Updated): This now contains only the regression tests for the nested bracket/code span bug.
   - **PR #779** (Updated): Rebased on top of #777.

2. Proptest seed explanation: the `proptest-regressions/.../block_builder.txt` seed was generated because the introduction of line numbers in `Block::Paragraph` (changed from `Vec<String>` to `Vec<(String, usize)>`) caused the `proptest_leading_blanks_do_not_alter_block_content` test to fail.

   When leading blank lines are fed into `BlockBuilder`, they shift the absolute line numbers of subsequent paragraph blocks. This caused the baseline run (without leading blanks) and the prefixed run (with leading blanks) to differ in their paragraph line numbers, failing the structural equality assertion.

   This was resolved by introducing a `finalize_test` helper in unit tests to normalize line numbers to `0` for comparisons. The seed generated during this failure has been checked in to preserve the regression guard. I also added a dedicated unit test `test_line_number_tracking` in `block_builder.rs` to explicitly assert that the line numbers are tracked correctly when not normalized.

3. Conflict resolution & cleanup (`helpers.rs` and `link_parser.rs`)
   - Rebased on top of `main` and resolved the conflict in `helpers.rs` by aligning with the new unified `strip_structural_prefixes_slice` helper.
   - For `link_parser.rs`, I reverted the temporary adjustment in `finalize_links_and_images` and instead extended the span for collapsed links inside the `broken_link_callback` using the existing `extend_collapsed_byte_end` helper, keeping the single forward pass intact as requested.

4. The following two cases fail (trigger MD013 violations) on current `main`:
   - Link with nested brackets: `-   [[T; N] builtin syntax]({url})`
   - Link with nested code and brackets: `-   [`[T; N]`builtin syntax]({url})`

   These are now explicitly covered by the new regression test in this PR (#777) and are resolved once the binary search optimization (PR #781) is applied, as it leverages `pulldown_cmark`'s robust link parsing instead of the legacy regexes (which fail on nested brackets).

rvben:
The tests are good and they cover real gaps. Holding only because this carries #781's commit, which regresses the standalone-link exemption for unbracketed destinations containing a space (details there). Goes in right behind #781 once that has a fallback, no changes needed here.

PR #779, opened 2026-08-01
fix(md013): prevent splitting nested links and allow fallback wrapping for soft spans

chandlerc:
• Parse split pieces of formatting spans to preserve nested atomic structures (e.g., preventing splitting of links nested inside emphasis).
• Introduce `is_hard` to `ElementSpan` to distinguish hard atomic elements (links, code, HTML) from soft ones (emphasis).
• Implement fallback in `break_before_attached` to allow breaking inside soft atomic spans if strict wrapping fails, preventing line length violations when they are adjacent to other unbreakable elements.
• Gate the fallback split retry so that atomic spans that fit within the line budget are not split by attached characters (such as punctuation).
• Update existing regression tests and add new ones for both cases.

### DIFF
--- a/src/lint_context/link_parser.rs
+++ b/src/lint_context/link_parser.rs
@@ -141,9 +141,10 @@ pub(super) fn parse_links_images_pulldown<'a>(
         content,
         options,
         Some(|link: BrokenLink<'_>| {
+            let span_end = extend_collapsed_byte_end(content, link.link_type, link.span.end);
             broken_links.push(BrokenLinkInfo {
                 reference: link.reference.to_string(),
-                span: link.span.clone(),
+                span: link.span.start..span_end,
                 link_type: link.link_type,
             });
             None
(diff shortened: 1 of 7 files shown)


chandlerc:
Note: depends on #777

rvben:
Two blockers here. It carries #781's commit, so it inherits the exemption regression described there.

And one from this commit: `break_before_attached` retries with `relax_soft_spans = true` without checking `options.atomic_spans` or the span's width, so an atomic span that fits gets split when an attached character pushes the line over. `**one two**,` with `reflow = true`, default `atomic-spans = true`, base vs `c6fd7c9f`:

| line-length | base | PR |
|---|---|---|
| 10 | `**one` / `two**,` | `**one` / `two**,` |
| 11 | `**one two**,` | **`**one` / `two**,`** |
| 12 | `**one two**,` | `**one two**,` |

Only at 11, where the span is 11 wide and fits but the `,` makes the line 12. At 10 both split it, correctly. That's the policy stated in the same file ("A span that alone exceeds the line budget is broken even when spans are atomic"), so the relaxed retry just needs the same gate the rest of the file uses: `!atomic_spans || element_len > line_length`.

chandlerc:
Should be fixed, and rebased to include the fix for #781

PR #780, opened 2026-08-01
fix(md013): track line numbers in BlockBuilder for list reflow

chandlerc:
First commit of the MD013 split stack. Tracks line numbers through BlockBuilder and list reflow to fix line exemptions inside lists.

Proptest seed update note:
- The change to `Block::Paragraph` to store line numbers `Vec<(String, usize)>` caused the `proptest_leading_blanks_do_not_alter_block_content` test to fail. This is because adding leading blank lines shifts the absolute line numbers of subsequent paragraph blocks, causing the baseline and prefixed block lists to differ in their line numbers.
- This was resolved by introducing a `finalize_test` helper in unit tests to normalize line numbers to 0 before comparison. The regression seed generated during this failure (`block_builder.txt`) has been checked in to preserve the case.

### DIFF
--- a/proptest-regressions/rules/md013_line_length/block_builder.txt
+++ b/proptest-regressions/rules/md013_line_length/block_builder.txt
@@ -1,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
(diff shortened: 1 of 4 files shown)


rvben:
Thanks @chandlerc, splitting the stack made this reviewable. Merging this one: self-contained, and a differential MD013 run against 7535e208 shows no behavior change from this commit alone.

Non-blocking note: `Block::Paragraph` now carries line numbers the list-item loop discards (`LineType::Content(content, _)`), so that fact lives in two places. They agree today (asserted across the suite, 11949/11949, with a wrong-offset control failing 171), just worth remembering if either side moves.

PR #781, opened 2026-08-01
perf(md013): optimize standalone link check using binary search on LintContext

chandlerc:
Replaces regex-based checks with O(log N) binary search over pre-parsed links in LintContext. 

This includes a fallback check for unbracketed link/image destinations containing spaces (e.g. Jekyll/Liquid template paths) which are not parsed as single links by pulldown-cmark. If the exact range lookup fails, we check for a partial parsed link starting at the line's beginning and verify that the remainder is a single parenthesized group.

**Note:** This drops the standalone link exemption for link-shaped lines inside fenced code blocks, as LintContext does not index links inside fenced code blocks.

### DIFF
--- a/src/lint_context/link_parser.rs
+++ b/src/lint_context/link_parser.rs
@@ -141,9 +141,10 @@ pub(super) fn parse_links_images_pulldown<'a>(
         content,
         options,
         Some(|link: BrokenLink<'_>| {
+            let span_end = extend_collapsed_byte_end(content, link.link_type, link.span.end);
             broken_links.push(BrokenLinkInfo {
                 reference: link.reference.to_string(),
-                span: link.span.clone(),
+                span: link.span.start..span_end,
                 link_type: link.link_type,
             });
             None
--- a/src/rules/md013_line_length.rs
+++ b/src/rules/md013_line_length.rs
@@ -503,7 +503,7 @@ impl Rule for MD013LineLength {
                 // Lines whose only content is a link/image are exempt.
                 // After stripping list markers, blockquote markers, and emphasis,
                 // if only a link or image remains, there is no way to shorten it.
-                if is_standalone_link_or_image_line(line) {
+                if is_standalone_link_or_image_line(ctx, line_number) {
                     continue;
                 }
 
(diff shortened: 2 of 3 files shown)


rvben:
The new lookup is a real improvement, confirmed: the nested-bracket and nested-code-span cases from #777 go 1 to 0 against 7535e208. But it regresses a case that passes today, so I can't merge it yet.

`has_matching_link_or_image` needs a `ctx.links` / `ctx.images` entry whose byte range is exactly the trimmed line. When pulldown-cmark parses no link spanning the whole line, the exemption just disappears. `line-length = 80`, MD013 counts, base vs `43b0a115`:

| input | len | base | PR |
|---|---|---|---|
| plain long prose (control) | 111 | 1 | 1 |
| `[label](https://example.com/...)` | 99 | 0 | 0 |
| `* [Front Matter Defaults]({{ '...' \| relative_url }})` | 92 | 0 | **1** |
| same line, destination in `<>` | 94 | 0 | 0 |

`<>` restores it and so does removing the interior spaces, so it's the exact-range check, not Liquid. Same for images under `stern = true, ignore-link-urls = false`: `Placeholder` 0 to 1. Under `reflow = true` it isn't just a report, `fmt` wraps the line.

So any Jekyll/Hugo/docs repo with a templated URL containing spaces gets new CI failures on files 0.2.48 passes. This commit is the base of the stack, so #777 and #779 have it too.

Needs a fallback rather than a rollback: when the exact-range lookup finds nothing, accept a line whose trimmed extent is a single link-shaped construct (or a parsed link whose start matches and whose end reaches the last `)`).

Also worth a line in the description: this drops the exemption for link-shaped lines inside fenced code blocks. Defensible, but it's a behavior change, and `docs/md045.md:25` here is one of them.

chandlerc:
Should be fixed (and rebased), but let me know if there are any other issues!

rvben:
Merged, along with #777 and #779. Thanks for working through the stack.

One follow-up landed on main afterwards (`722fad69`): the spaces-in-destination
fallback also matched a complete link followed by a parenthetical, so lines like

```
- [ripgrep](https://github.com/x/rg) (a line-oriented search tool that recursively searches here)
```

became exempt. Two conditions tell the cases apart. A spaced destination leaves
the link text parsed as `LinkType::Shortcut` with an empty url, while a complete
link is `Inline` with its own url; and CommonMark forbids whitespace between `]`
and `(`, so the helper now uses `trim_end()` instead of `trim()`. Everything the
PR set out to exempt still is.

PR #786, opened 2026-08-03
feat(md077): support fixed continuation indent config

### LINKED ISSUE #784: [MD077] Use MkDocs indentation even in Commonmark documents
### Problem or motivation

I use `rumdl` alongside [`editorconfig-checker`](https://editorconfig-checker.github.io), which is a tool that enforces that documents follow the [`.editorconfig`](https://editorconfig.org) configuration.

When applied to `.md` documents with `indent = 2`, this means that second lines must be indented 4 spaces, then 6, etc. My documents are GFM documents, and to pass the `editorconfig-checker` check, next lines must be 4 spaces. However `MD077` instructs Commonmark documents to use 3 spaces. If it were an MkDocs document, it could be 4. But there does not seem to be a way to instruct `rumdl` to use 4 spaces intentionally.

This causes a conflict during the linting/fixing stage of my development pipeline with tools arguing with each other.

### Proposed solution

* Let `.editorconfig` override what MD077 validates against.
* Allow Commonmark documents to use MkDocs indentation rules.
* Allow the user to somehow specify $2n$, $3n$, $4n$, …, etc., where $n$ is the indentation level.

### Alternatives considered

Disable MD077.

### Feature category

Improve existing rule

### Example markdown

This is what is expected by `rumdl` with 2-space indentation and a second line for a Commonmark document.

```markdown
* This is a sentence in a list item.

   This is a second sentence in a list item.

```

This is what I'd like to achieve as a valid configuration.

```markdown
* This is a sentence in a list item.

    This is a second sentence in a list item.
```

### Contribution

### Additional context

I haven't learned Rust yet. Sorry.

DavidPandleton:
Adds an optional `indent` setting to MD077 so documents that follow a different indentation convention (4 spaces for MkDocs, or a `.editorconfig`-driven style) can pin the required continuation indent instead of the content-column-derived default. When set, the requirement becomes marker column + indent, per item.

Closes #784.

### DIFF
--- a/docs/md077.md
+++ b/docs/md077.md
@@ -234,7 +234,40 @@ to the content column.
 #               column (mdformat parity). Tight under-indented lazy
 #               continuation is additionally flagged and snapped up.
 style = "any"
+
+# Fixed continuation indent relative to each item's marker, e.g. 4 for
+# MkDocs-style documents. When set, this overrides the content-column-derived
+# requirement (including the MkDocs flavor's 4-space minimum). Unset (None)
+# keeps the default content-column behavior.
+# indent = 4
 ```
+
+### `indent`
+
+By default MD077 derives the required continuation indent from each item's
+content column (the W+N rule; under the MkDocs flavor a 4-space minimum is
+enforced). When a fixed `indent` is configured, the requirement becomes the
+item's *marker column plus the configured indent*, which lets documents that
+follow a `.editorconfig` style (e.g. `indent = 2`, or MkDocs's 4-space
+convention) stay consistent with tools like `editorconfig-checker`:
+
+```toml
+[MD077]
+indent = 4
+```
+
+With `indent = 4`, a top-level item (marker at column 0) requires continuation
+content at 4 spaces, and a nested item (marker at column 2) requires it at 6:
+
+```markdown
+- a long line
+    that continues 4 spaces past the marker
+  - nested item
+      that continues 4 spaces past the nested marker
+```
+
+Continuation content below the fixed requirement is flagged and fixed up to it
+(the same rules for blank-line escapes and over-indentation still apply).
 
 ### `style = "aligned"`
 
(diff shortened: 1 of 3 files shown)


rvben:
Thanks for this, and for the clean implementation. The per-item `marker column + indent` shape is right, and I checked it against the document in #784: with `indent = 4`, `rumdl fmt` produces exactly the 4-space form that was asked for, and a re-check is clean.

I merged this and pushed four corrections on top rather than sending you round-trips, since most of them need repo internals that are not reasonable to expect from outside:

1. **The config key validator did not know about `indent`.** `impl_rule_config_methods!(MD077Config)` derives a rule's known keys by serializing the default config through JSON to TOML, which drops `None` fields, so an `Option` key is invisible to it. `[MD077] indent = 4` printed `Unknown option for rule MD077: indent` and failed the run under `--deny-config-warnings`, even though the setting itself worked. The `nullable` arm of that macro fixes it. Worth knowing for any future `Option` config key.
2. **Under `--flavor mkdocs`, a configured indent could lower the 4-space minimum.** With `indent = 2`, a document correctly indented at 4 was reported as over-indented and `fix` rewrote it to 2, which Python-Markdown no longer reads as part of the list item. That minimum is a compatibility floor rather than a style default, so a configured indent may now raise it but not lower it. It stays scoped to that flavor: under `standard`, `indent = 2` still means 2.
3. **One message claimed a consequence that was not true.** A configured indent produced "needs 4 spaces of indentation to remain part of the list" for content that was still part of the list. A requirement coming from the configured indent now says "to match the configured indent"; the structural and MkDocs wordings are unchanged.
4. **`indent = 0` is now rejected.** It asks for continuation at the marker's own column, which ends the list item, so correctly indented content was flagged as "over-indented (expected 0, found 2)".

I also corrected `docs/md077.md`, which said content below the requirement is flagged under the same rules as the default. That is not true of tight lazy continuation: `style` still decides which lines get checked, and `indent` only decides what they are checked against.

Thanks again.

PR #794, opened 2026-08-04
fix(md018): accept the tags option instead of reporting it as unknown

shuvamk:
`[MD018] tags` is documented in `docs/md018.md` and works, but the config key validator reports it as unknown, so a run with `--deny-config-warnings` fails on a valid config. On `main` @ `26543648`:

```console
$ printf '[MD018]\ntags = true\n' > .rumdl.toml
$ printf '#todo something\n\n# Real heading\n' > doc.md
$ rumdl check --no-cache --deny-config-warnings doc.md
[config warning] Unknown option for rule MD018: tags

Success: No issues found in 1 file (26ms)
$ echo $?
2
```

"No issues found" is the setting doing its job — `#todo` is skipped. Only the validator cannot see the key.

## Cause

The same one you fixed for MD077 in `cba175c0`: the plain `impl_rule_config_methods!` arm derives the known keys by serializing the default config, which drops `None` fields.

Here the `nullable` arm alone is not enough. `tags` also carried `skip_serializing_if = "Option::is_none"`, so the key is gone from the serialized value before `config_schema_table` can substitute a sentinel for it. Reverting either line on its own leaves the warning in place. MD018 is the only rule config using that attribute.

## Trade-off

As with MD077's `indent`, `rumdl config` now prints the nullable sentinel for `tags` — 11 keys across 8 rules already do on `main`. `rumdl init` and `rumdl config --output json` are byte-identical. Happy to filter the sentinel out of the config display path in a separate PR if you want that cleaned up.

## Tests

One test in `tests/config/documented_alias_reachability_test.rs` beside the MD077 one: `tags = true` produces no warning and takes effect, and `tagz = true` is still reported as unknown. Verified it fails with the source change stashed and passes with it restored.

Full suite, `cargo fmt --check` and `rumdl schema check` green on the branch.

### DIFF
--- a/src/rules/md018_no_missing_space_atx.rs
+++ b/src/rules/md018_no_missing_space_atx.rs
@@ -385,7 +385,7 @@ impl Rule for MD018NoMissingSpaceAtx {
         self
     }
 
-    crate::impl_rule_config_methods!(MD018Config);
+    crate::impl_rule_config_methods!(MD018Config, nullable);
 }
 
 #[cfg(test)]
--- a/src/rules/md018_no_missing_space_atx/md018_config.rs
+++ b/src/rules/md018_no_missing_space_atx/md018_config.rs
@@ -16,7 +16,7 @@ pub struct MD018Config {
     /// When true, single-hash patterns like `#tag`, `#project/active` are
     /// skipped. When null/unset, defaults to true for Obsidian flavor
     /// and false otherwise.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
     pub tags: Option<bool>,
 }
 
--- a/tests/config/documented_alias_reachability_test.rs
+++ b/tests/config/documented_alias_reachability_test.rs
@@ -124,6 +124,27 @@ fn md077_indent_is_not_reported_as_an_unknown_option() {
 }
 
 #[test]
+fn md018_tags_is_not_reported_as_an_unknown_option() {
+    const TAG_LINE: &str = "# T\n\n#todo something\n";
+
+    let (baseline, _) = findings_for("MD018", "[MD018]\n", TAG_LINE);
+    assert_eq!(baseline, 1, "control: `#todo` is flagged without the setting");
+
+    let (count, stderr) = findings_for("MD018", "[MD018]\ntags = true\n", TAG_LINE);
+    assert_eq!(count, 0, "control: the setting must actually take effect");
+    assert!(
+        !stderr.contains("Unknown option"),
+        "a supported setting must not be reported as unknown, got:\n{stderr}"
+    );
+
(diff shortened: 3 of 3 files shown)


DavidPandleton:
Nice catch on both halves of this. I hit the same class of bug on MD077 in #786, where the config key validator also rejected a valid Option key. The `nullable` macro arm is exactly the fix rvben applied there.

The `skip_serializing_if = "Option::is_none"` removal is the part worth calling out: that attribute is what drops the field during the JSON->TOML serialization the validator uses to derive known keys, so without removing it the nullable arm alone would not have made the key visible. Good that both changes landed together.

The typo case in the test (`tagz`) is a nice addition too, it proves the validator still catches real mistakes.

rvben:
Fixed on `main` and released in [v0.2.51](https://github.com/rvben/rumdl/releases/tag/v0.2.51), by a general fix rather than this one, so I am closing the PR. Your diagnosis is what found it, and `8ec7bbab` cites this report.

Your reading of the cause was right, including the part about the `nullable` arm alone not being enough. Rather than adjust MD018's config so `tags` survives serialization, validation stopped reading `default_config_section()` altogether: that table is a listing of defaults, and a key whose default cannot be written down has no business being in it. Validation now reads a separate `Rule::config_schema()` derived from the same struct `from_config` deserializes, so accepted keys cannot drift from honored ones. It was a class rather than a one-off, and 16 rule files changed.

That also removes the trade-off you flagged. Sentinels contain a NUL byte and are unwritable as TOML, and four properties are now pinned against each other: every published default must be accepted by the validator, every schema key must survive a real load-and-validate, no user-facing table may carry a sentinel, and `rumdl config --defaults` must produce a config rumdl itself accepts. So nothing is printed for `tags`, for MD077's `indent`, or for the other nine.

Verified on v0.2.51, with the control showing the validator is strict rather than silent:

```console
$ printf '[MD018]\ntags = true\n' > .rumdl.toml
$ rumdl check --no-cache --deny-config-warnings doc.md
Success: No issues found in 1 file (10ms)          # exit 0

$ printf '[MD018]\ntagz = true\n' > .rumdl.toml
$ rumdl check --no-cache --deny-config-warnings doc.md
[config warning] Unknown option for rule MD018: tagz (did you mean: tags?)
doc.md:1:2: [MD018] No space after # in heading [*]  # exit 2
```

One part the general fix did not cover is your test, so I have taken it verbatim: `md018_tags_is_not_reported_as_an_unknown_option` now sits in `tests/config/documented_alias_reachability_test.rs` next to the MD077 case from #786. The class has a general guard, but nothing pinned your instance of it, and the `tagz` control is what makes it worth having.

Thank you for the report. The `--deny-config-warnings` exit code was the detail that made this obviously a defect rather than a preference, and it is why the fix went wide instead of touching one rule.

PR #803, opened 2026-08-08
Mention MegaLinter integration in README

nvuillam:
Hi! rumdl is now shipped out of the box in [MegaLinter](https://megalinter.io/), so I added a small mention in the CI/CD Integration section :) If anything looks off on https://megalinter.io/latest/descriptors/markdown_rumdl/ just let me know. Thanks for the great tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ With intelligent caching, subsequent runs are even faster - rumdl only re-lints 
   - [GitHub Actions](#github-actions)
     - [Inputs](#inputs)
     - [Examples](#examples)
+  - [MegaLinter](#megalinter)
 - [Rules](#rules)
 - [Flavors](#flavors)
   - [Supported Flavors](#supported-flavors)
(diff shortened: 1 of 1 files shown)


rvben:
Thanks, and good to have this documented from your side.

One small thing I fixed on the way in rather than sending you back for another round: the new paragraph is a single 232 character line, and rumdl's own `.rumdl.toml` sets MD013 `line-length = 200`, so `rumdl check README.md` flagged it here. `rumdl fmt README.md` wraps it, which is what I ran.

I also added a MegaLinter section to the docs site (`docs/usage/ci-cd.md`), since that is where most people look and a README-only mention leaves it out. It shows:

```yaml
MARKDOWN_DEFAULT_STYLE: rumdl
ENABLE_LINTERS:
  - MARKDOWN_RUMDL
```

taken from your descriptor page's variable table, which notes `MARKDOWN_DEFAULT_STYLE` has to be `rumdl` for the linter to be active. If that is not the shape you would recommend to users, say so and I will change it.

nvuillam:
@rvben thanks for accepting the PR !

there is an internal discussion about what should be the default markdown formatter in MegaLinter ^^

For the moment, markdownlint is still the most popular, but keep the good work and it will be rumdl someday :)

rvben:
Thanks @nvuillam! I'll continue to do my best to make rumdl the default choice for every markdown user 😉

PR #805, opened 2026-08-09
fix(md075): ignore Obsidian wikilink aliases in prose after tables

### LINKED ISSUE #804: MD075: ignore Obsidian wikilink aliases when detecting table rows
## Reproduction

With the `obsidian` flavour enabled:

```markdown
| Existing | Table |
| --- | --- |

See [[Target Note|display label]].
```

## Expected behaviour

No MD075 diagnostic. The blank line ends the table, and the pipe belongs to the valid Obsidian wikilink alias.

## Actual behaviour

MD075 reports an orphaned table row. The same parser mistake also reports a non-table paragraph containing an aliased wikilink as a pipe-formatted row without a header. `rumdl check --fix` must not be used for this case: it can absorb the prose into the preceding table and split it at the alias pipe.

## Proposed fix

When the flavour is `obsidian`, treat pipes inside `[[target|alias]]` as non-delimiters while testing whether a line is a candidate table row. This should be scoped to candidate-row detection so genuine orphaned table rows continue to be reported and other flavours retain their current behaviour.

I have a tested fork branch with this targeted change, but it needs rebasing onto current upstream before a clean PR can be opened.

krisb:
## Summary

- prevent MD075 from treating a prose paragraph containing an Obsidian aliased wikilink as an orphaned table row
- treat an aliased wikilink as one table cell for MD056 in the Obsidian flavour
- document the supported Obsidian table behaviour

The changes are scoped to the Obsidian flavour. Genuine orphaned table rows remain detectable, and other flavours retain GFM behaviour.

Fixes #804

### DIFF
--- a/docs/flavors.md
+++ b/docs/flavors.md
@@ -10,7 +10,7 @@ rumdl supports multiple Markdown flavors to accommodate different documentation 
 | [gfm](flavors/gfm.md)                   | GitHub Flavored Markdown             | MD033, MD034                                                                              |
 | [mkdocs](flavors/mkdocs.md)             | MkDocs / Material for MkDocs         | MD024, MD031, MD033, MD038, MD040, MD042, MD046, MD049, MD050, MD052, MD056               |
 | [mdx](flavors/mdx.md)                   | MDX (JSX in Markdown)                | MD013, MD033, MD037, MD039, MD044, MD049                                                  |
-| [obsidian](flavors/obsidian.md)         | Obsidian knowledge base              | MD011, MD012, MD018, MD028, MD033, MD034, MD037, MD038, MD044, MD049, MD061, MD064, MD069 |
+| [obsidian](flavors/obsidian.md)         | Obsidian knowledge base              | MD011, MD012, MD018, MD028, MD033, MD034, MD037, MD038, MD044, MD049, MD056, MD061, MD064, MD069 |
 | [pandoc](flavors/pandoc.md)             | Pandoc Markdown                      | MD022, MD029, MD031, MD032, MD034, MD037, MD040, MD042, MD051, MD052                      |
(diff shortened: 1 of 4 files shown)


rvben:
Merged, thanks. The wikilink pipe really is part of the link rather than a cell divider, and rumdl was reading it the GFM way in a flavor where that is wrong.

Two follow-ups landed on top of your change:

**Quoted brackets and a blank target are prose** (`a8aed70e`). Masking the pipe inside `[[...]]` also fired on two shapes that are not links, and masking them merged cells that were already correct, so a well formed table got a column-count complaint:

```markdown
| `[[` | mid | `]]` |
| [[ | ]] | z |
```

Code binds tighter than a link, so brackets quoted as code open nothing, and the half before the pipe names the note, so with nothing there the brackets are prose that happens to straddle a divider. Both now split exactly as GFM does. The masker and the wikilink scan share one definition of an inline code span, so they cannot disagree about where code begins.

**A table's extent is found in the document's flavor** (`ad139af7`). Splitting a row into cells honoured the flavor, but the walk deciding which lines belong to a table did not, so the original problem still had a path through it:

```markdown
| A | B |
| - | - |
| x | y |
[[Foo|bar]] is a note about the rabbit.
```

With no blank line the paragraph was absorbed as a table row: MD055 and MD056 fired on it, and `rumdl fmt` rewrote it to `| [[Foo|bar]] is a note about the rabbit. |  |`. It now reports MD058 and inserts the missing blank line. The same threading fixed MD013, where a paragraph holding a wikilink was read as a table row and so exempt from the line-length check and from reflow, and MD075's headerless-group column count, which admitted rows flavor-aware and then counted them flavor-blind.

Under `standard` and `mkdocs` nothing changes: the pipe is still a cell divider everywhere it was before.

This ships in the next patch release.

Issue #126, opened 2025-10-30
MD028 flags consecutive quotes


zane:
## Reproduction steps

```
rumdl check - <<EOF
# Heading

> Hello.

> Goodbye.
EOF
```

## Expected results

```
No issues found in <stdin>
```

## Actual results

```
<stdin>:4:1: [MD028] Blank line inside blockquote (level 1) [*]

Found 1 issue(s) in <stdin>
```

rvben:
Hi @zane,

Thanks for raising this issue. While this is valid according to CommonMark, but as also the [markdownlint docs](https://github.com/DavidAnson/markdownlint/blob/main/doc/md028.md) explain; "Some Markdown parsers will treat two blockquotes separated by one or more blank lines as the same blockquote, while others will treat them as separate blockquotes." I have updated the [rumdl docs for MD028](https://github.com/rvben/rumdl/blob/main/docs/md028.md) to reflect and explain this better.

That means that there are 3 options:

**Option 1:** Add the `>` marker on the blank line to make it one continuous blockquote:
```markdown
> Hello.
>
> Goodbye.
```

**Option 2:** Add separating text to clarify they're distinct blockquotes:
```markdown
> Hello.

And then:

> Goodbye.
```

**Option 3:** Disable MD028 if you prefer allowing this pattern:
```bash
rumdl check --disable MD028 yourfile.md
```
(Or of course, disable it in the config or with inline comments.)

serl:
Hello, just adding a little use case we had, which might me interesting (or not). Note that disabling the rule is what we did and we're totally fine with that:

```markdown
> [!TIP]
> Here's a github tip

> [!NOTE]
> Here's a github note
```

They must be kept separated, otherwise the first will "eat" the second:

> [!TIP]
> Here's a github tip

> [!NOTE]
> Here's a github note

```markdown
> [!TIP]
> This tip eats the note below
> [!NOTE]
> What am I becoming?
```

> [!TIP]
> This tip eats the note below
> [!NOTE]
> What am I becoming?


(image FTR)

<img width="258" height="297" alt="Image" src="https://github.com/user-attachments/assets/75790a3c-49ac-483e-9e28-86dd1b47f8ec" />

PR #806, opened 2026-08-09
fix(md028): respect distinct blockquotes in standard flavors

krisb:
## Summary

Make MD028 flavour-aware for ordinary blockquotes:

- accept blank-line-separated blockquotes in `standard` (including GFM/CommonMark aliases) and `obsidian`, where the syntax has defined separate-blockquote semantics
- retain the existing strict portability check and opt-in merge for other flavours
- document the distinction and add regression coverage for the runtime behaviour, formatter interactions, and character ranges

## Relationship to #126

This implements a flavour-aware resolution to the trade-off discussed in [#126](https://github.com/rvben/rumdl/issues/126). It does not remove MD028’s conservative, backwards-compatible behaviour: users targeting flavours without a defined blockquote boundary retain the current warning and opt-in merge.

For `standard` (including GFM/CommonMark aliases) and `obsidian`, a blank line is defined to end one blockquote and begin another. In those modern flavours, reporting the pair marks valid, intentionally separate source and annotation blocks as a violation.

## Validation

- `cargo fmt --check`
- `cargo test`
- `rumdl fmt --check docs/`
- CLI matrix: `standard` and `obsidian` clean; `mkdocs` reports MD028 for the same input


rvben:
Thanks for putting this together. I am going to decline it, and I would rather
say so now than after another round.

The change makes `is_blank_inside_blockquote` return false whenever the flavor
is Standard or Obsidian, which is the default. That makes MD028 inert by
default: most tests that expected a standard-flavor finding are re-pointed at
MkDocs, and the two that stay on Standard now assert the rule is silent. The
rule exists because renderers disagree about a blank line
between two `>` blocks: CommonMark ends the quote there, Python-Markdown (so
MkDocs) merges the two, and a document that moves between them changes meaning.
That disagreement is not settled by which flavor a project declares, it is
settled by where the document is rendered, and the same file is routinely
rendered by more than one. So the rule stays default-on; that is #126, closed
2025-11-01, and the GFM alert case that motivated #126 is already handled by
the alert recognition inside the rule.

If MD028 is wrong for your documents, `[global] disable = ["MD028"]` turns it
off and the docs say so. If there is a specific blockquote shape it misreads,
open that as an issue with the input and I will look at the rule.

Issue #760, opened 2026-07-26
MD013 cannot normalize italicized text followed by a number


denialhaag:
### rumdl version

0.2.43

### Installation method

Pre-built binary

### Bug description

This one is admittedly very niche, but MD013 fails to normalize the specific example below.

### Steps to reproduce

Run `rumdl check` on the following file:

```markdown
_I cannot italicize text that is exactly 80 characters long followed by 123456_, 123456.
```

`rumdl check` fails with:

```
example.md:1:81: [MD013] Line length 88 exceeds 80 characters
```

### Expected behavior

The line is broken after the comma, or no error is eported.

### Actual behavior

An error is reported.

### Configuration

```toml
[MD013]
line-length = 80
code-blocks = false
headings = false
reflow = true
reflow-mode = "normalize"
```

### Operating system

macOS

### Additional context

Both of these examples are formatted as expected (i.e., the line is broken):

```markdown
_I cannot italicize text that is exactly 80 characters long followed by 123456_, word.
```

```markdown
_I cannot italicize text that is exactly 80 characters long followed by 12345678_, 12345678.
```

denialhaag:
The problem seems to have existed since version 0.2.33. The versions before break the line after the comma.

rvben:
Fixed in v0.2.44.

Reflow refused to break a line before any `<up to 9 digits>.` or `<up to 9 digits>)` token, to avoid turning prose into an ordered list. That guard was too broad: a list can only interrupt a paragraph when it is numbered 1 and its first item has content, so a bare `123456.` is prose to the parser and wrapping before it cannot change the document. That is why `word.` and your other variants wrapped and this one did not, and why you got the unfixable "Line length" message instead of the fixable normalize one.

Your input now gives:

```markdown
_I cannot italicize text that is exactly 80 characters long followed by 123456_,
123456.
```

The regression tests use the parser as the oracle: they render the original paragraph and the reflowed output and require the two to match, so a wrap that changes the document's structure fails no matter which line it lands on.

Thanks for the bisect and for the working variants. Those controls are what showed the guard was firing on the wrong thing.

Issue #748, opened 2026-07-21
Issue with `atomic-spans`


denialhaag:
### rumdl version

0.2.37

### Installation method

Pre-built binary

### Bug description

MD013 may fail for spans exceeding the configured line length even though `atomic-spans` is enabled.

### Steps to reproduce

Run `rumdl` on the following file:

```markdown
_This is a very, very, very, very, very, very, very long line that exceeds 80 characters with some `code` inside._
```

`rumdl` fails with:

```markdown
example.md:1:81: [MD013] Line length 114 exceeds 80 characters
```

### Expected behavior

It shouldn't fail because `atomic-spans` is enabled.

### Actual behavior

It does fail.

### Configuration

```toml
[MD013]
line-length = 80
atomic-spans = trie
code-blocks = false
headings = false
reflow = true
reflow-mode = "normalize"
```

### Operating system

macOS

### Additional context

From what I've seen, MD013 only fails if two "span types" are mixed (e.g., "_" and "`").

rvben:
Fixed in [v0.2.39](https://github.com/rvben/rumdl/releases/tag/v0.2.39).

Thanks for the precise report - your observation that it only fails when two span types are mixed was exactly right, and it generalizes: any nested `` ` ``, `*`, `_` or `~` inside the outer span triggered it.

`atomic-spans = true` never meant "never break an emphasis span". A span that on its own exceeds `line-length` is still broken, because holding it whole leaves a line that can never fit. That escape hatch was being cancelled for any content containing an inline marker, so a span with nested markup was unbreakable at any length - MD013 then reported a violation its own fixer refused to touch. Worse, normalize mode re-joined a correctly hand-wrapped paragraph back into one over-long line, so `rumdl fmt` produced output that `rumdl check` rejected. This was a regression in 0.2.37; 0.2.36 handled your input correctly.

Such a span now wraps at whitespace outside any nested construct. Nested code spans, links, images and emphasis are each held as one unbreakable unit, so a break can never collapse a code span's literal whitespace or a link destination's structural whitespace. That last part also fixed an older bug where `[text](<a b c>)` inside an emphasis span could be split mid-destination.

denialhaag:
Thanks a lot for the quick fix, @rvben! 🙂 

One small observation: This example still fails, likely for a related reason:

```markdown
***This is a very, very, very, very, very, very, very, very, very, very long line that is bold and italic.***
```

rvben:
[...]

PR #812, opened 2026-08-12
fix(md013): end a sentence-per-line at a sentence that closes inside emphasis

### LINKED ISSUE #811: MD013 sentence-per-line: a sentence ending inside emphasis is joined onto the next, producing an unfixable line-length violation
**rumdl version:** 0.2.54 (also reproduced on every release from 0.2.45 onward)
**Installation method:** Pre-built binary

### Bug description

In `reflow-mode = "sentence-per-line"`, a sentence whose terminator sits inside an emphasis span is not recognised as finished. The span is joined onto the sentence that follows it, and the joined line exceeds `line-length` — so `rumdl fmt` turns conforming input into an `MD013` violation that `rumdl fmt` cannot then fix.

This is the same shape as #748 and #760: `fmt` produces output that `check` rejects.

### Steps to reproduce

`.rumdl.toml`:

```toml
[MD013]
line-length = 120
reflow = true
reflow-mode = "sentence-per-line"
```

`test.md` — three prose lines, none of them over `line-length`:

```markdown
# T

**An enum column may not be an atom a keyset scan resumes from.**
`ORDER BY` would follow the declaration index while the keyset predicate followed the value.
The two disagree, and the page skips rows.
```

```console
$ rumdl check test.md
test.md:3:1: [MD013] Paragraph should have one sentence per line (found 2 sentences across 3 lines) [*]

$ rumdl fmt test.md
test.md:3:1: [MD013] Paragraph should have one sentence per line (found 2 sentences across 3 lines) [fixed]

$ rumdl check test.md
test.md:3:121: [MD013] Line length 158 exceeds 120 characters

$ rumdl fmt test.md
test.md:3:121: [MD013] Line length 158 exceeds 120 characters
```

### Expected behavior

The input is already one sentence per line, so `check` reports nothing and `fmt` leaves it alone. This is what 0.2.44 does.

### Actual behavior

`check` counts 2 sentences where there are 3. `fmt` joins lines 3 and 4 into a single 158-character line, then reports it as a line-length violation with no fix available — `rumdl rule MD013` advertises `Fix: Fix is always available`. The rewrite is not idempotent and leaves the file worse than it found it.

The blast radius is not limited to one file. On a 49-file corpus that 0.2.43 reports clean, 0.2.54 reports 13 findings, rewrites 8 files, and leaves 5 unfixable `MD013` line-length findings behind. Every one of the 13 is a bold lead-in ending `.**` followed by a line opening with a code span or a lowercase word, which is an ordinary Markdown idiom rather than a corner case.

### Regression range

Bisected over the released binaries: 0.2.44 leaves the reproducer's longest line at 92 characters, and 0.2.45 and every release since produce 158.

The change is `207823b664ea5ca34c2c7c1ec5b2335366ea5b17`, "fix(md013): keep every word when a sentence boundary falls inside emphasis". It removed `handle_emphasis_sentence_split`, which tested the emphasis **content** for a terminator, and folded emphasis into the ordinary text path, which tests the **rendered span**. The rendered form of `**… resumes from.**` ends in `*`, not `.`, so `ends_with_sentence_punct` at `src/utils/text_reflow.rs:905` stopped firing for the single-sentence fast path in `reflow_elements_sentence_per_line`.

The plain-text analogue still works, which is why this only shows up with emphasis: for `Plain lead sentence.` followed by a code span the parser emits a text element ending in `.`, the fast path fires, and the line is emitted. An emphasis span is the one sentence-ending shape that fails the test, purely because of its closing marker.

Italic and strikethrough are affected identically:

```markdown
*Ital lead.*
`Code` next.
Tail here.
```

### Configuration

As above. `require-sentence-capital` is left at its default `true`; setting it to `false` masks the symptom but trades it for split-after-abbreviation false positives, so it is not a workaround.

I have a fix and tests ready and will open a PR against this issue.

no-simpler:
## Description

Fixes #811.

In `reflow-mode = "sentence-per-line"`, a sentence whose terminator sits inside an emphasis span was read as unfinished and absorbed the sentence following it. The joined line ran past `line-length`, which MD013 then reported as a line-length violation carrying no fix — `fmt` produced output `check` rejects, and a second `fmt` could not repair it. Same shape as #748 and #760.

```markdown
**An enum column may not be an atom a keyset scan resumes from.**
`ORDER BY` would follow the declaration index while the keyset predicate followed the value.
The two disagree, and the page skips rows.
```

Rewritten after review. The first attempt changed the single-sentence fast path in `reflow_elements_sentence_per_line`; this one leaves that path alone and puts the whole rule in `is_sentence_boundary`, which is what both the counter and the reflow read.

### Why the rule belongs there

MD013 decides sentence boundaries twice. `src/rules/md013_line_length.rs` counts with `split_into_sentences` to produce "found N sentences across M lines", and `reflow_elements_sentence_per_line` calls `split_into_sentences_with_set` per element to decide where lines end. A rule that moves one without the other makes `fmt` produce output `check` still rejects — the same non-convergence #811 reports. Both funnel into `is_sentence_boundary`, so a rule placed there moves them together.

### The three commits

**`fix(md013): let a code span open a sentence`.** `require-sentence-capital` reads the first character after the terminator and wants an uppercase letter. A code span starts on a backtick, so there is no case to read, and the gate held the sentence open. `!` and `?` already accept any following character; a period now accepts a code span on the same terms. The abbreviation, decimal and initial guards all run first, and the rule carries one more of its own — not after a digit, so `Steps: 1. \`init\` the repo, 2. \`build\` it.` stays one sentence.

**`fix(md013): end a sentence at the span that opened it`.** A sentence that both opens and closes inside one span is delimited by that span's own markers, so the word after it decides nothing. A span closing *mid*-sentence is the bolded-command idiom, and there the case of the following word is the only thing marking it as one sentence. The relaxation is therefore narrow: the run of markers closing the sentence has to be the run the sentence opened with, marker for marker, and that marker may not appear in between. The run is read backward from the space every boundary arm already requires, rather than forward from the terminator, because the quoted arms leave the markers two or more characters past it.

The sentence start is carried as an `Option`, absent where the caller cannot vouch for one. The element-wise caller emits on trailing punctuation alone in its single-sentence path, so it can emit where the splitter would have kept the sentence open, and nothing after such an emit opens a sentence.

**`fix(md013): split a sentence-per-line tail that ends on an atomic element`.** Found while testing the first two. A code span, link or autolink is appended to the line being built without the splitter ever reading it, on the understanding that a later text element re-splits the line. The last element of a paragraph has no later one, so this input made `check` count two sentences that `fmt` would not break:

```markdown
**An enum column may not be an atom that a keyset scan resumes from, and the planner cannot fix that.**
`ORDER BY key, id`
```

The tail is now split once at the final flush. A tail carrying a non-breaking space keeps the whole-line path, since the splitter trims each sentence with `str::trim`, which counts one as whitespace.

### The must-fix, and why there is no guard for it

The suggested `followed_by_source_whitespace` is unreachable once the rule lives in `is_sentence_boundary`. Every ASCII arm of that function already binds the index of a required space — `pos + 1` for the plain case, `pos + 2` through `pos + 4` for the quoted and emphasized ones, and `chars.get(end_pos) == Some(&' ')` for the footnote arm — and the consume loop takes exactly the markers that arm skipped plus that space. No break can land where the source had no whitespace, so the roughly 650 spans with no whitespace after them are untouched.

That claim is about the ASCII path only. The CJK arm returns before any of it and requires no following whitespace at all, so `**中文。**（括号）` does break where the source had none. Pre-existing, and neither new rule reaches it, so it is left alone here rather than folded in without its own measurement.

### Behaviour this does not change

A boundary *inside* a span reaches the multi-sentence branch, so what `207823b6` added is preserved: `**First. Second.**` still becomes `**First.\nSecond.**` rather than closing and reopening its markers. Nested emphasis closes on a run that differs from the one the sentence opened with, so `**Bold with *ital.* inside**` is one sentence. Every idiom named in review is unchanged, pinned by test:

```markdown
Click **Save.** then restart.
Select the Explorer view in the Activity Bar, and select the **New File...** button to create a file.
The **code .** command opened VS Code in the current working folder.
**Tip:** Start with **Dev Containers: Add Dev Container Configuration Files...** in the Command Palette.
```

The last one is why the marker may not appear between the two runs. Without that condition a `**Note:**` lead-in pairs its own opening markers with the closing markers of a later span, which puts a slice of the 635 mid-sentence spans straight back.

## Type of Change

Bug fix (non-breaking change which fixes an issue).

## Testing

- `tests/regressions/md013_reflow_span_terminated_sentence_issue_811_test.rs` — end-to-end through the real binary, over every shape #811 lists plus a quoted span and a trailing code span. Asserts the invariant rather than the exact break: content preservation, idempotence, and that `fmt` output is clean under `check`, both for conforming input and for the same paragraphs written on one line.
- `tests/utils/text_reflow_test.rs` — engine-level coverage of each rule, its guards, the mid-sentence idioms above, and the trailing-atomic case including its non-breaking-space exception.
- `cargo nextest run --profile dev`: 12571 passed, 0 failed. `cargo test --doc`, `cargo fmt --check` and `cargo clippy --workspace --lib --bins --tests -- -D warnings -D clippy::uninlined_format_args` all clean.
- Measurements per commit are in the review reply.

## Checklist

- [x] Tests added covering the fix
- [x] Full test suite passes
- [x] `cargo fmt` and `cargo clippy` clean
- [x] Conventional commits referencing the issue
- [x] `docs/md013.md` documents both new exceptions to `require-sentence-capital`
- [x] `CHANGELOG.md` left alone — it is generated at release time


rvben (changes requested):
Thanks for this, and for the clear write-up in #811. The construct is real and rumdl handles it wrong today. I ran the change against a local corpus of about 90,000 Markdown files, and I think the rule needs a different shape before it lands.

## Must Fix

**`text_reflow.rs`: the fast path now ends a line where the source had no whitespace.** A line break renders as a space, so breaking after `**Bold.**text` inserts a space the author did not write. The element after a span carries the gap, so the guard is cheap:

```rust
/// Whether the source separated the element at `idx` from what follows it.
///
/// The element after a span carries the gap: text keeps its own leading
/// whitespace, and any other element abutting the span had none. A line may only
/// end where the source already had whitespace, since the break renders as a
/// space.
fn followed_by_source_whitespace(elements: &[Element], idx: usize) -> bool {
    match elements.get(idx + 1) {
        None => true,
        Some(Element::Text(text)) => text.starts_with(char::is_whitespace),
        Some(_) => false,
    }
}
```

applied at the boundary check:

```rust
let closed = trim_closing_markers(trimmed);
let may_end_line = closed.len() == trimmed.len() || followed_by_source_whitespace(elements, idx);
if may_end_line
    && ends_with_sentence_punct(closed)
    && !text_ends_with_abbreviation(closed, &abbreviations)
```

There are about 650 spans in the corpus with no whitespace after them.

## Should Fix

**The rule bypasses `require-sentence-capital`, and that is where most of its cost is.** rumdl has two sentence-boundary computations: the element fast path this PR extends, which never consults `require-sentence-capital`, and the text path, which does. Making a closing span an unconditional terminator means an emphasized span ends a sentence in cases where the identical plain text would not.

Measured on the corpus, the change fires on 223 spans that genuinely begin their own sentence (what #811 is about) and on 635 that sit mid-sentence. The mid-sentence ones are dominated by the bolded UI command idiom:

```markdown
Start with **Dev Containers: Add Dev Container Configuration Files...** in the Command Palette.
Select the Explorer view in the Activity Bar, and select the **New File...** button to create a file.
The **code .** command opened VS Code in the current working folder.
```

Each of those is one sentence, and with this PR `rumdl fmt` cuts it after the closing `**`.

The narrower change that fixes #811's code-span cases with no measured regression is to leave the boundary rule alone and teach the sentence-start gate that a code span opens a sentence, since its delimiter is a backtick rather than a letter and the case of what it holds belongs to the code:

```rust
if first_char == '`' {
    return true;
}
```

That covers the code-span half of the 13 cases in #811. The lowercase half needs more: splitting there is right only when the emphasized span began its own sentence, because `Click **Save.** Then restart.` is the common shape where only the closing word is emphasized and the follower gate is what keeps it correct.

Happy to take this as two changes if you would rather split it, or to land the gate condition separately and keep this PR focused on the lowercase case.

## Positive

- The commit message argues the rule change explicitly rather than just asserting the fix, which is what made it possible to test the property directly.
- `trim_closing_markers` is the right helper and is reusable regardless of which boundary rule wins.
- #811 lists all 13 real occurrences with their shapes. That corpus is what made the measurement above possible.

no-simpler:
Both findings taken, and the measurement changed my mind about where the rule belongs. Rewritten as three commits, force-pushed.

## Where the rule moved, and why

The first attempt changed the single-sentence fast path only. That was wrong for a reason your review points at without naming: MD013 decides sentence boundaries twice, and the two have to agree. `src/rules/md013_line_length.rs` counts with `split_into_sentences` at nineteen call sites to produce "found N sentences across M lines"; `reflow_elements_sentence_per_line` calls `split_into_sentences_with_set` per element to decide where lines end. Changing the reflow alone means the counter keeps reporting a paragraph the rewriter has already rewritten, and the finding never clears — the same non-convergence #811 reports.

So the fast path is back to what it was, `trim_closing_markers` is gone, and both rules live in `is_sentence_boundary`.

## Must fix

Unreachable once the rule is there, so there is no guard for it. Every ASCII arm of `is_sentence_boundary` already binds the index of a required space — `pos + 1` plain, `pos + 2` through `pos + 4` quoted and emphasized, `chars.get(end_pos) == Some(&' ')` for footnotes — and the consume loop takes exactly the markers that arm skipped plus that space. The ~650 spans with no whitespace after them are untouched, and `**Bold.**text` is pinned by test.

One correction to that claim, which I would rather state than have you find: it holds for the ASCII path only. The CJK arm returns before all of it and requires no following whitespace at all, so `**中文。**（括号）` does break where the source had none, and `preserves_content` waves it through because it only forbids *removing* a break. That is pre-existing and neither new rule reaches it, so I left it out rather than fold an unmeasured change into this. Happy to open it separately if you want it.

## Should fix

Your narrow alternative is commit 1, as written, plus one conjunct: not after a digit. The decimal guard only fires when a digit follows the period too, so without it `Steps: 1. \`init\` the repo, 2. \`build\` it.` splits at every item.

The lowercase half is commit 2, under your rule — split only when the emphasized span began its own sentence. Two things the obvious form of that gets wrong, both found by testing rather than by reasoning:

**Read the run backward from the space, not forward from the terminator.** The quoted arms put the markers two or more characters past it, so a forward scan misses `**"An enum column may not be an atom a keyset scan resumes from."**` — one of #811's own shapes, still broken.

**The marker may not appear between the two runs.** Matching "same character, same length" alone is not "this run closes the span that opened the sentence". `**Tip:** Start with **Dev Containers: Add Dev Container Configuration Files...** in the Command Palette.` pairs the lead-in's own `**` with the closing `**` of a later span and splits after `Files...**` — a `**Note:**` / `**Tip:**` / `**Step 1:**` lead-in is common enough that this alone would put a slice of your 635 back. With the condition it is one sentence.

The cost is conservatism: a literal `*` or `~` inside the span suppresses the split. `**The glob \*.md matches.** npm reads it.` stays one sentence. That seemed the right direction to be wrong in.

The sentence start is carried as an `Option`, absent where the caller cannot vouch for one. The element-wise caller's single-sentence path emits on trailing punctuation alone, so it emits in places the splitter would not have, and a span closing after such an emit opened nothing. Without that, `Ship it.` followed by `**bold lead.** npm follows.` splits one sentence across three lines.

## A third one, which the review did not name

Commit 3. A code span, link or autolink is appended to the line being built without the splitter ever reading it, on the understanding that a later text element re-splits the line. The last element of a paragraph has no later one:

```markdown
**An enum column may not be an atom that a keyset scan resumes from, and the planner cannot fix that.**
`ORDER BY key, id`
```

Before commit 3, `check` counted two sentences here and `fmt` joined the two lines into 122 characters instead of breaking them. Pre-existing, but commit 1 is what makes the counter see the second sentence, so it belongs in this branch. The tail is split once at the final flush; a tail carrying a non-breaking space keeps the whole-line path, since the splitter's `str::trim` counts one as whitespace and the edge trimming there exists to keep it.

## Measurements

Two corpora, no 90,000-file corpus of my own — please re-measure. Counts are MD013 findings before any `fmt`, so they are the counter's view, which is the half your review is about.

**rumdl's own 130 Markdown files**, with `reflow = true`, `reflow-mode = "sentence-per-line"`, `line-length = 120` forced on:

| build | findings |
| --- | --- |
| `main` (0049eb2d) | 790 |
| + commit 1, code span opens a sentence | 800 |
| + commit 2, span-opened sentence | 802 |
| + commit 3, trailing atomic tail | 802 |

Commit 2 moves two paragraphs in the whole tree; commit 3 does not move the counter at all, only the rewrite. Every new finding I sampled is a real second sentence in your own documentation:

```markdown
Images are not checked. `![alt](test.md)` is not a link a reader follows, and
Only a leading `~/` (or a bare `~`) is a home reference. `~` is a legal
- a variation selector that follows a base character. `U+26A0 U+FE0F` renders as the
```

`README.md:801` is the one that shows the point. Today it reports `Line length 200 exceeds 120 characters` with no fix attached; on the branch it reports two sentences on one line, which `fmt` fixes.

**A 98-file corpus written under this mode**, the one #811 came out of:

| build | findings | files `fmt` rewrites | findings left after `fmt` |
| --- | --- | --- | --- |
| 0.2.43 | 0 | 0 | 0 |
| 0.2.55 | 22 | 11 | 10, all unfixable length findings |
| branch | 3 | 3 | 2 |

The two left are genuine over-long sentences, 156 and 235 characters, which the mode reports with no fix so a human shortens them — not lines `fmt` created. A second `fmt` changes nothing.

## Deliberately not here

- The three arms accepting `.* `, `.** `, `.__ ` and `.~~ ` miss single-tilde `~Strike.~`, triple `***Emphasis.***` and mixed nesting `**_Bold ital._**`. Reading a whole marker run would close that, but it widens the set of runs commit 2's predicate compares, so the two want measuring together rather than in one branch. Opened separately as #816.
- The CJK arm, above.
- `reflow_elements_semantic`'s `prev_ends_at_sentence` strips trailing quotes and brackets but not `* _ ~`, so a line ending `**Bold.**` is not read as ending a sentence there either. Same family, different mode.

## If you would rather have it exact

Commit 2's predicate reads marker runs out of the source. The exact version is to swap `extract_code_spans` in `split_into_sentences_with_set` for `extract_emphasis_and_code_spans`, which returns both from one pulldown pass, and ask whether an `EmphasisSpan` starts at the sentence's first non-whitespace character and ends at the required space. Nesting-proof, no false negatives, and it drops the conservatism above. I did not take it because `extract_code_spans` skips the parser entirely when there is no backtick while the combined extractor parses on any of `* _ ~` too, and `split_into_sentences` is called nineteen times over in md013. If you would rather pay that, or want it computed lazily on first need, say so and I will switch it.

Also happy to split the branch if you would rather take commit 1 on its own first — it stands alone and carries most of the measurement.

rvben (changes requested):
Re-measured the branch at e8bb7e06 against `main`, same forced config as the
first review (`MD013.reflow=true`, `reflow-mode='sentence-per-line'`,
`line-length=120`), over the ~90,000-file corpus, and then a second pass of
targeted inputs against each of the three commits.

The good news first. 2,547 files come out different from `main`; every diff I
sampled is the split #811 asks for, both binaries converge (a second `fmt`
changes nothing on either), and the #811 reproducer is fixed: `main` joins the
bold sentence onto the code-span line and leaves a 158-column line it cannot
fix, the branch leaves the three lines alone and a second `check` is clean. The
idioms from the first review hold too. Both boundary computations funnelling
through `is_sentence_boundary` is the right frame and I will keep it.

Three things do not hold, each reproduced on the branch and not on `main`, so
they are the branch's. Inputs and outputs verbatim, `rumdl fmt` with the config
above and `MD041` disabled:

1. Commit 3's final flush splits inside a trailing link, image alt text, angle
   destination or title. Input:

   ```markdown
   Opening sentence. [First. Second](https://example.com)
   ```

   Branch:

   ```markdown
   Opening sentence.
   [First.
   Second](https://example.com)
   ```

   Same for `Opening sentence. First. Second` and for
   `Opening sentence. [target](https://example.com "Title. More")`, and the
   angle-bracket destination comes out worse: the branch turns
   `Opening sentence. [target](<https://example.com/First. Second>)` into
   `[target](<<https://example.com/First>.` / `Second>)`, so the destination is
   rewritten, not just broken. The flush hands raw Markdown to
   `split_into_sentences`, which knew only code spans as unbreakable. That
   was `main`'s gap as much as yours, and it is closed in v0.2.56 (72871b98
   and 78bafa98, on top of aabb4209): the splitter takes every atomic inline
   range from the same nested-structure walk the reflow uses, so a boundary
   never falls inside a link, image, wikilink, math span or HTML tag, in the
   check and in the fix alike, and it reads sentence openers off that parse
   too. I rebased your three commits onto v0.2.56 to check: all four
   inputs above come out whole through the flush, and 2 and 3 below still
   reproduce, so those are the branch's alone. Please add the four to your
   tests so the flush pins it, since it is the one path that reaches the
   splitter with raw text. Expect conflicts in `text_reflow.rs` on your second
   commit, all in the same area: `span_opened_sentence` lands next to `main`'s
   `link_opener_len` (keep both); `split_into_sentences_with_set` now takes
   the document's reference definitions, so it needs both that and your
   `opens_sentence`; `is_sentence_boundary` now takes a `SentenceText` view,
   so your `sentence_start` goes on the end of that; and `main`'s
   `closes_before_next` probe calls the splitter and needs your `opens`
   argument, as a closure parameter, since the loop assigns `opens_sentence`
   while the closure lives. Your third commit's final flush and your new tests then call
   `split_into_sentences_with_set` / `split_into_sentences` with the old
   arity, so pass `options.defined_references.as_ref()` there and `None` in
   the tests. `split_into_sentences_custom` is gone on `main`. With that
   resolution the branch builds and every reflow test on both sides passes.

2. Commit 1's code-span-opens-a-sentence arm accepts the last period of an
   ellipsis. Input:

   ```markdown
   Choose File... `Open Folder` appears in the menu.
   ```

   `main` leaves it alone. Branch:

   ```markdown
   Choose File...
   `Open Folder` appears in the menu.
   ```

   The arm needs the same ellipsis guard the prose path already has (a period
   preceded by a period is not a sentence end).

3. Commit 2 takes an opening emphasis span for a complete sentence whenever
   its markers balance. Input:

   ```markdown
   **Figure 1.** shows the architecture.
   ```

   `main` leaves it alone. Branch:

   ```markdown
   **Figure 1.**
   shows the architecture.
   ```

   Same for `**1.** first step initializes the repository.` and
   `*New File...* button creates a file.` A label at the start of a line
   closes its markers and ends in punctuation exactly like a one-sentence
   span does. The discriminator is the one the prose path already applies:
   the text after the closing marker has to start a sentence, a capital after
   the space. `main` splits `**Note.** Then a capital continues.` today and
   leaves `**Note.** then lowercase continues.` alone; the branch splits both.
   Routing the emphasis arm through the same capital guard fixes all three
   inputs and keeps the #811 case.

The CJK link-text case I named last time is `main`'s and is covered in
v0.2.56 (72871b98) as well (`功能。[指南。](url)更多。` now keeps the link whole); the
`...` before a code span through the trailing-punctuation emit path is still
`main`'s and stays mine.

On the exact variant: no, the parse-on-any-marker cost across the nineteen
call sites is not worth the conservatism it removes, and the conservative
predicate errs in the safe direction. #816 after this lands.

Requesting changes for 2 and 3, and a rebase for 1. Thanks for the rewrite and
the measurements; this is close.

in-style:
Rebased onto `main` at `84890191`, three commits again. Findings 1 and 2 are taken as written. Finding 3 is taken, but not by the route you named — the discriminator turned out to be something else, and chasing it changed commit 1 too.

## 1. The rebase, and the flush

Your resolution held in every particular, including the closure parameter: `closes_before_next` calls the splitter while the loop assigns `opens_sentence`, so it takes the claim as an argument rather than capturing it. `split_into_sentences_with_set` takes `options.defined_references.as_ref()` at the flush and the tests pass `None`.

All four of your inputs come out whole through the flush now, and all four are pinned in `test_sentence_per_line_trailing_link_is_never_split_inside`. Agreed that this is the one path reaching the splitter with raw Markdown, so it is the one that needs them.

## 2. The ellipsis guard does not exist

One correction before the fix, since it changes what the fix can lean on. I looked for the guard in `text_reflow.rs`, in `md013_line_length.rs`, across `src/` by code search, and through the commit history. The only ellipsis-aware site in the tree is `src/utils/sentence_utils.rs:211`, inside MD064's backward-looking `is_after_sentence_ending`, and it reads the other way:

```rust
// Check for ellipsis (...) - always a valid sentence ending
if idx >= 2 && chars[idx - 1] == '.' && chars[idx - 2] == '.' {
    return true;
}
```

What leaves your input alone on `main` is not a guard but `opens_sentence_in_strict_mode`: a backtick is not uppercase, numeric or CJK. Take the following word out of the picture and `main` splits an ellipsis today —

```markdown
Choose File... Open Folder appears in the menu.
```

— which `main` breaks after `File...`.

So the branch introduces the rule rather than reusing one, and scopes it to the arm. Widening it to every period arm would fix the input above as well, but it would also put `text_reflow` in direct contradiction with `sentence_utils` over the same three characters, and move your corpus in a change you did not ask for. Left out; happy to open it separately with a measurement if you want the two modules reconciled.

## 3. Labels, and why the capital guard cannot be the answer

Routing the emphasis arm through `require-sentence-capital` does fix all three of your inputs. It also takes the whole of commit 2 with it. `main` already splits `**Note.** Then a capital continues.` and already holds `**Note.** then lowercase continues.`, so once the arm defers to that gate there is nothing left for a marker-reading rule to decide. The only case it was deciding is the one the gate refuses.

That case is half of #811. The issue's thirteen occurrences are "a bold lead-in ending `.**` followed by a line opening with a code span **or a lowercase word**", and the reporter on the issue thread describes the same split in their own corpus. Commit 1 covers the code-span half. Dropping commit 2 leaves the other half joined, which I checked rather than assumed: with commit 2 out, `**An enum column may not be an atom a keyset scan resumes from.**` followed by `npm would follow the declaration index.` comes back as one line.

What actually separates your three inputs from #811's shapes is the form of the period, not the marker runs and not the following word:

| input | period follows | verdict |
| --- | --- | --- |
| `**Figure 1.** shows the architecture.` | a digit | not a terminator |
| `**1.** first step initializes the repository.` | a digit | not a terminator |
| `*New File...* button creates a file.` | a period | not a terminator |
| `**An enum column … resumes from.** npm follows.` | a letter | a terminator |

So commit 2 keeps reading the marker runs, and takes only a period commit 1 found terminal. A label ending in a digit or an elision cannot reach it. `Click **Save.** then restart.`, `The **code .** command …`, `**Tip:** Start with **Dev Containers: … Files...** …` and `**Bold with *ital.* inside** and more.` are all still one sentence each, and are pinned alongside your three.

## 4. What measurement then changed in commit 1

The digit conjunct as I wrote it last round was too broad, and the corpus said so. Three of the six findings the rebased branch left on my corpus were this:

```markdown
- **A2.**
  `UserProvider::refreshUser` re-mints cookie tokens via auth-service on every session refresh.
  Token issuance policy is auth-service's.
```

`main` joins the first two lines into 129 characters and then reports the length with no fix. So does 0.2.62, and so did my branch, because `A2.` ends in a digit. But `A2.` is not an enumerator — it is a label whose last character happens to be a digit, and this is one of the shapes #811 was filed over. 0.2.43 leaves it alone.

An enumerator's period stands bare. A label's sits between a span's closing markers and the space. That is the whole distinction, and it is enough:

```rust
let elision = pos > 0 && chars[pos - 1] == '.';
let digit_run = pos > 0 && chars[pos - 1].is_numeric();
let bare = space_pos == pos + 1;

if first_char == '`' && !elision && !(digit_run && bare) {
    return true;
}
```

`Steps: 1. \`init\` the repo, 2. \`build\` it.` and `Pinned at 0.2.43. \`rumdl\` reads it.` stay one sentence; `**A2.** \`UserProvider\` re-mints the token.` becomes two. Commit 2 does not get the `bare` relaxation, because nothing there vouches for the text after the markers the way a code span vouches for itself — that is what keeps `**1.** first step …` whole.

One divergence from 0.2.43 that I would rather name than have you find. Written as one line, `**1.** \`init\` the repo.` now splits after the label, where 0.2.43 leaves it. Written as two lines it goes the other way, and in the branch's favour: 0.2.43 and the branch both leave them alone, and `main` joins them. So that shape is one more #811 instance rather than a cost.

## Measurements

Findings are MD013 counts before any `fmt`, so they are the counter's view.

**rumdl's own 135 Markdown files**, with `reflow = true`, `reflow-mode = "sentence-per-line"`, `line-length = 120` forced on:

| build | findings | of those, length findings with no fix |
| --- | --- | --- |
| `main` | 959 | 166 |
| branch | 975 | 164 |

The two that clear are the ones your first review pointed at. `docs/flavors.md:60` reports `Line length 188 exceeds 120 characters` on `main` and two sentences on the branch, which `fmt` fixes; `README.md:827` does the same at 200 characters. Every added finding I sampled is a real second sentence in your documentation. `fmt` converges: the second pass changes no file.

**A 95-file corpus written under this mode**, the one #811 came out of:

| build | findings | files `fmt` rewrites | findings left after `fmt` |
| --- | --- | --- | --- |
| 0.2.43 | 0 | 0 | 0 |
| 0.2.62 | 16 | 10 | 8, all unfixable length findings |
| branch | 3 | 3 | 2 |

The two left are genuine over-long sentences, 156 and 235 characters, which the mode reports with no fix so a human shortens them — not lines `fmt` created. A second `fmt` changes nothing.

Against your four regression inputs and your four flush inputs, the branch and `main` produce byte-identical output. The branch differs from `main` on exactly five inputs, all of them #811's shapes.

`cargo test`: 13,203 pass. The only failures are `perf::rule_complexity_regression_test`, which fail on unmodified `main` on this machine too, with a different arm each run.

## Deliberately not here

- #816, as agreed, after this lands.
- The class-wide elision question above, which needs `sentence_utils` reconciled with it rather than a second rule pointing the other way.
- The CJK arm, unchanged.

no-simpler:
Rebased onto current `main`, three commits again. Findings 1 and 2 are taken as written. Finding 3 is taken, but not by the route you named — the discriminator turned out to be something else, and chasing it changed commit 1 too.

## 1. The rebase, and the flush

Your resolution held in every particular, including the closure parameter: `closes_before_next` calls the splitter while the loop assigns `opens_sentence`, so it takes the claim as an argument rather than capturing it. `split_into_sentences_with_set` takes `options.defined_references.as_ref()` at the flush and the tests pass `None`.

All four of your inputs come out whole through the flush now, and all four are pinned in `test_sentence_per_line_trailing_link_is_never_split_inside`. Agreed that this is the one path reaching the splitter with raw Markdown, so it is the one that needs them.

## 2. The ellipsis guard does not exist

One correction before the fix, since it changes what the fix can lean on. I looked for the guard in `text_reflow.rs`, in `md013_line_length.rs`, across `src/` by code search, and through the commit history. The only ellipsis-aware site in the tree is `src/utils/sentence_utils.rs:211`, inside MD064's backward-looking `is_after_sentence_ending`, and it reads the other way:

```rust
// Check for ellipsis (...) - always a valid sentence ending
if idx >= 2 && chars[idx - 1] == '.' && chars[idx - 2] == '.' {
    return true;
}
```

What leaves your input alone on `main` is not a guard but `opens_sentence_in_strict_mode`: a backtick is not uppercase, numeric or CJK. Take the following word out of the picture and `main` splits an ellipsis today —

```markdown
Choose File... Open Folder appears in the menu.
```

— which `main` breaks after `File...`.

So the branch introduces the rule rather than reusing one, and scopes it to the arm. Widening it to every period arm would fix the input above as well, but it would also put `text_reflow` in direct contradiction with `sentence_utils` over the same three characters, and move your corpus in a change you did not ask for. Left out; happy to open it separately with a measurement if you want the two modules reconciled.

## 3. Labels, and why the capital guard cannot be the answer

Routing the emphasis arm through `require-sentence-capital` does fix all three of your inputs. It also takes the whole of commit 2 with it. `main` already splits `**Note.** Then a capital continues.` and already holds `**Note.** then lowercase continues.`, so once the arm defers to that gate there is nothing left for a marker-reading rule to decide. The only case it was deciding is the one the gate refuses.

That case is half of #811. The issue's thirteen occurrences are "a bold lead-in ending `.**` followed by a line opening with a code span **or a lowercase word**", and the reporter on the issue thread describes the same split in their own corpus. Commit 1 covers the code-span half. Dropping commit 2 leaves the other half joined, which I checked rather than assumed: with commit 2 out, `**An enum column may not be an atom a keyset scan resumes from.**` followed by `npm would follow the declaration index.` comes back as one line.

What actually separates your three inputs from #811's shapes is the form of the period, not the marker runs and not the following word:

| input | period follows | verdict |
| --- | --- | --- |
| `**Figure 1.** shows the architecture.` | a digit | not a terminator |
| `**1.** first step initializes the repository.` | a digit | not a terminator |
| `*New File...* button creates a file.` | a period | not a terminator |
| `**An enum column … resumes from.** npm follows.` | a letter | a terminator |

So commit 2 keeps reading the marker runs, and takes only a period commit 1 found terminal. A label ending in a digit or an elision cannot reach it. `Click **Save.** then restart.`, `The **code .** command …`, `**Tip:** Start with **Dev Containers: … Files...** …` and `**Bold with *ital.* inside** and more.` are all still one sentence each, and are pinned alongside your three.

## 4. What measurement then changed in commit 1

The digit conjunct as I wrote it last round was too broad, and the corpus said so. Three of the six findings the rebased branch left on my corpus were this:

```markdown
- **A2.**
  `UserProvider::refreshUser` re-mints cookie tokens via auth-service on every session refresh.
  Token issuance policy is auth-service's.
```

`main` joins the first two lines into 129 characters and then reports the length with no fix. So does 0.2.62, and so did my branch, because `A2.` ends in a digit. But `A2.` is not an enumerator — it is a label whose last character happens to be a digit, and this is one of the shapes #811 was filed over. 0.2.43 leaves it alone.

An enumerator's period stands bare. A label's sits between a span's closing markers and the space. That is the whole distinction, and it is enough:

```rust
let elision = pos > 0 && chars[pos - 1] == '.';
let digit_run = pos > 0 && chars[pos - 1].is_numeric();
let bare = space_pos == pos + 1;

if first_char == '`' && !elision && !(digit_run && bare) {
    return true;
}
```

`Steps: 1. \`init\` the repo, 2. \`build\` it.` and `Pinned at 0.2.43. \`rumdl\` reads it.` stay one sentence; `**A2.** \`UserProvider\` re-mints the token.` becomes two. Commit 2 does not get the `bare` relaxation, because nothing there vouches for the text after the markers the way a code span vouches for itself — that is what keeps `**1.** first step …` whole.

One divergence from 0.2.43 that I would rather name than have you find. Written as one line, `**1.** \`init\` the repo.` now splits after the label, where 0.2.43 leaves it. Written as two lines it goes the other way, and in the branch's favour: 0.2.43 and the branch both leave them alone, and `main` joins them. So that shape is one more #811 instance rather than a cost.

## Measurements

Findings are MD013 counts before any `fmt`, so they are the counter's view.

**rumdl's own 135 Markdown files**, with `reflow = true`, `reflow-mode = "sentence-per-line"`, `line-length = 120` forced on:

| build | findings | of those, length findings with no fix |
| --- | --- | --- |
| `main` | 959 | 166 |
| branch | 975 | 164 |

The two that clear are the ones your first review pointed at. `docs/flavors.md:60` reports `Line length 188 exceeds 120 characters` on `main` and two sentences on the branch, which `fmt` fixes; `README.md:827` does the same at 200 characters. Every added finding I sampled is a real second sentence in your documentation. `fmt` converges: the second pass changes no file.

**A 95-file corpus written under this mode**, the one #811 came out of:

| build | findings | files `fmt` rewrites | findings left after `fmt` |
| --- | --- | --- | --- |
| 0.2.43 | 0 | 0 | 0 |
| 0.2.62 | 16 | 10 | 8, all unfixable length findings |
| branch | 3 | 3 | 2 |

The two left are genuine over-long sentences, 156 and 235 characters, which the mode reports with no fix so a human shortens them — not lines `fmt` created. A second `fmt` changes nothing.

Against your four regression inputs and your four flush inputs, the branch and `main` produce byte-identical output. The branch differs from `main` on exactly five inputs, all of them #811's shapes.

`cargo test`: 13,203 pass. The only failures are `perf::rule_complexity_regression_test`, which fail on unmodified `main` on this machine too, with a different arm each run.

## Deliberately not here

- #816, as agreed, after this lands.
- The class-wide elision question above, which needs `sentence_utils` reconciled with it rather than a second rule pointing the other way.
- The CJK arm, unchanged.

rvben:
Thanks for the careful rebase and write-up. This is much closer. The atomic-tail changes look good to me now, including the link, image, title, and angle-destination cases.

I ran the current head (`5982c427`) locally. All 12,973 nextest tests passed, and formatting, doctests, and strict clippy are clean.

I found two cases that I think still need fixing.

First, the period heuristic handles the examples from the previous review, but not labels in general:

```markdown
**A.** first step initializes the repository.
**Example.** shows the architecture.
```

The branch reports both as two sentences and moves the text after the label onto a new line. Current `main` leaves them alone.

It also misses the reverse case:

```markdown
**The process returned 0.** npm retries it.
```

That stays joined because every digit-ending span is treated as a label.

I don’t think marker balance and the character before the period are enough to distinguish a complete emphasized sentence from a label. That is the ambiguity `require-sentence-capital` currently handles.

The second case is smaller:

```markdown
First phrase. `literal backtick continues the same sentence.
```

This gets split because the new condition treats any raw backtick as a code-span opener. An unmatched backtick is ordinary Markdown text, so this should check that the position actually opens a parsed code span.

I suggest narrowing the PR a little:

- Keep the trailing atomic-element flush.
- Keep the code-span fix, but check the parsed code-span ranges.
- Leave out the general `span_is_a_sentence` override for now.

We can handle the ambiguous lowercase case separately. One option is to preserve an existing source line break without inferring the same boundary from joined text. I’m also happy to pick up that part locally if you would rather keep this PR focused.

Thanks for working through all the review rounds. The main regression is important, and most of the difficult plumbing looks solid now.

no-simpler:
Both cases reproduce, both are yours, and I have taken the narrowing exactly as you laid it out. Two commits now: the code-span rule and the atomic-tail flush.

## The labels

Your counterexamples are decisive, and they break the rule in both directions at once:

```markdown
**A.** first step initializes the repository.
**Example.** shows the architecture.
**The process returned 0.** npm retries it.
```

The first two split on the branch because nothing about `A.` or `Example.` says "label" to a rule reading marker runs. The third stays joined because the only thing my rule had to go on was the digit, and a sentence is perfectly free to end on one. I was treating the character before the period as a proxy for a distinction it cannot carry. You are right that this is what `require-sentence-capital` is for.

`span_is_a_sentence` is gone, and with it `span_opened_sentence`, the `sentence_start` parameter, the `opens_sentence` claim threaded through `reflow_elements_sentence_per_line`, and the closure parameter on `closes_before_next`. All three inputs now come out exactly as `main` produces them.

I should correct something I argued last round. I read #811's "a code span **or a lowercase word**" as evidence that the lowercase half carried real weight, and used that to keep the rule. Measured against the corpus the issue came from, dropping it costs nothing: the branch reports the same 3 findings, rewrites the same 3 files, and leaves the same 2 behind as it did with the span rule in place. What was actually doing that work is the bare-versus-label distinction on a digit's period in the first commit, not the span rule. So the narrowing is free here, and I should have measured that before defending it.

## The backtick

Fixed by reading the parse instead of the character. `NestedStructure` already walks the document for `atomic`; it now keeps the code spans it finds there as their own list, and the arm asks whether a code span opens at that offset:

```rust
fn opens_code_span(&self, pos: usize) -> bool {
    self.char_offsets
        .get(pos)
        .is_some_and(|&start| self.code_spans.binary_search_by_key(&start, |&(s, _)| s).is_ok())
}
```

`First phrase. \`literal backtick continues the same sentence.` is one sentence again, and pinned.

## What the first commit still carries

The period-form guard stays, since it is what the code-span arm needs once it steps around the capital gate: one mark of an elision never terminates, and a period closing a digit run terminates only where it stands bare. Bare it is an enumerator or a version — `Steps: 1. \`init\` the repo, 2. \`build\` it.` and `Pinned at 0.2.43. \`rumdl\` reads it.` stay whole. Between a span's closing markers and the space it is a label's, so `**A2.** \`UserProvider\` re-mints the token.` is two sentences. That shape is what three of my corpus's findings actually were, and 0.2.43 leaves them alone.

Note that this reads only the *code-span* arm. It never decides a boundary the capital gate would have refused on its own for any other follower, which is what kept your three inputs safe.

## Measurements

**rumdl's own 135 Markdown files**, `reflow = true`, `reflow-mode = "sentence-per-line"`, `line-length = 120` forced:

| build | findings | of those, length findings with no fix |
| --- | --- | --- |
| `main` (0d5bbfcc) | 958 | 166 |
| branch | 972 | 164 |

The two that clear are still `docs/flavors.md:60` (188 characters, no fix) and `README.md:827` (200). `fmt` converges — the second pass changes no file.

**The 95-file corpus #811 came out of:**

| build | findings | files `fmt` rewrites | findings left after `fmt` |
| --- | --- | --- | --- |
| 0.2.43 | 0 | 0 | 0 |
| 0.2.62 / `main` | 16 | 10 | 8, all unfixable length findings |
| branch | 3 | 3 | 2 |

The 2 left are genuine over-long sentences, 156 and 235 characters, reported with no fix so a human shortens them. A second `fmt` changes nothing.

Over a 25-input probe set the branch differs from `main` on four, all of them a code span opening the line after a span. Two are the fix, restoring what 0.2.43 does — the reported reproducer, and the `**A2.**` list item, both of which arrive already one sentence per line and stay that way:

```markdown
- **A2.**
  `UserProvider::refreshUser` re-mints cookie tokens via auth-service on every session refresh.
```

The other two are the same shapes written inline, where the counter now sees a boundary 0.2.43 did not: `**A2.** \`UserProvider\` re-mints the token.` and `**1.** \`init\` the repo.` split, where 0.2.43 leaves each on one line. Nothing is joined either way, so `fmt` regresses nothing; it does mean a corpus that writes such lead-ins inline picks up findings, and that is part of the +14 on your own documentation above. I would rather name it than have it turn up in your corpus run.

Every non-`perf` test passes. The only failures here are in `perf::rule_complexity_regression_test`, which are load-dependent on this machine and reproduce on unmodified `main` with a different arm each run; they pass for you, so treat mine as noise.

## The lowercase half

Your framing of it is better than mine, and I think it is the right shape for a separate change: preserve an existing source line break rather than infer the same boundary out of joined text. The defect #811 reports is `fmt` *joining* lines that were already one sentence each — recovering a boundary from the joined form is a harder problem than the bug requires, and it is the part that keeps producing false positives.

Please do pick it up locally if you would rather own it; you have the corpus to measure it against and I would only be guessing at the shapes outside mine. If you would rather I took it, say so and I will open it separately once this lands, together with #816.

Thanks for the three rounds of measurement on this. The regression tests here now cover only the code-span half, which is what this pull request claims.

rvben (approved):
Thanks for narrowing the fix. I re-reviewed `a02dc75`, and the two remaining sentence-boundary false positives are now addressed without broadening the heuristic again.

The prior label/continuation cases remain unchanged, parsed code spans are distinguished from unmatched backticks, and the trailing atomic-element case now converges cleanly between `fmt` and `check`.

I also verified the focused regressions, full unit suite, doctests, formatting, and clippy locally. No remaining blocking findings from my review. Approved.

PR #815, opened 2026-08-13
docs(nav): add Code Block Tools page to navigation

eread:
To help with discoverability of the Code Block Tools feature (https://github.com/rvben/rumdl/blob/4997ff07b63c7a54d6a495ebcd3ce7f23ee1173f/docs/code-block-tools.md), let's add it to the website navigation.


eread:
@rvben Could you review? I realise the feature is "in preview", but increased visibility might mean increased usage which might mean increased stability over time?

WDYT?

rvben:
@eread, good idea! Else it might be in preview forever and that's also not what was the plan, of course.

Thanks for this. 🚀

PR #817, opened 2026-08-14
fix(links): read backslash escapes in a reference definition label

### LINKED ISSUE #814: MD034 false positive for reference definitions with an escaped closing bracket in link labels
### rumdl version

0.2.54

### Installation method

mise

### Bug description

MD034 (URL without angle brackets) is incorrectly reported for a reference definition with a link label containing an escaped closing bracket.

CommonMark spec says:

> A [link label](https://spec.commonmark.org/0.31.2/#link-label) begins with a left bracket ([) and ends with the first right bracket (]) that is not backslash-escaped. Between these brackets there must be at least one character that is not a space, tab, or line ending. Unescaped square bracket characters are not allowed inside the opening and closing square brackets of [link labels](https://spec.commonmark.org/0.31.2/#link-label). A link label can have at most 999 characters inside the square brackets.

(<https://spec.commonmark.org/0.31.2/#link-label>)

Escaped square brackets are allowed inside link labels. In particular, the CommonMark spec includes an example with an escaped opening bracket ([Example 549]). However, a link label containing an escaped closing bracket appears to be handled incorrectly by rumdl.

[Example 549]: https://spec.commonmark.org/0.31.2/#example-549

### Steps to reproduce

1. Create a file `test.md` with this content:

   ```markdown
   # Reference Example

   * [this is a link to example.com][ref1\[\]]
   * [this is also a link to example.com][ref2\[]
   * [this is also a link to example.com][ref3]

   [ref1\[\]]: https://example.com/ref1
   [ref2\[]: https://example.com/ref2
   [ref3]: https://example.com/ref3
   ```

2. Run `rumdl check test.md`
3. See the output message

### Expected behavior

No warnings are emitted

### Actual behavior

MD034 is incorrectly reported for the first reference definition.

```
test.md:7:13: [MD034] URL without angle brackets or link formatting: 'https://example.com/ref1' [*]

Issues: Found 1 issues in 1 file (8ms)
Run `rumdl fmt` to automatically fix 1 of the 1 issues
```

### Configuration

```toml
No configuration
```

### Operating system

Linux

### Additional context

_No response_

hdimer:
## Description

Fixes #814.

`REF_DEF_PATTERN` reads a reference definition's label as `[^\]]+`, which stops at the first `]` on the line whether or not it is escaped. CommonMark §6.3 ends a label at the first `]` that is **not** backslash-escaped, so `[ref1\[\]]: https://example.com/ref1` has the label `ref1\[\]` — but the regex commits to the `\]` as the terminator, then finds `]` where it needs `:`, and the line matches nothing at all. It is not a reference definition as far as `ctx.reference_defs` is concerned, so MD034 reads the rest of the line as ordinary paragraph text and reports its destination as a bare URL:

```
test.md:7:13: [MD034] URL without angle brackets or link formatting: 'https://example.com/ref1' [*]
```

The reporter's `[ref2\[]` case works today only because an escaped `[` needs no escape handling to get past — that label still ends at a real `]`. That the two forms of the same construct disagree is what makes this read as an oversight rather than a decision.

The label group now takes the escape alternation the destination and title groups have carried all along:

```
[^\]]+   ->   (?:[^\]\\]|\\.)+
```

The comment above the regex already documents why the angle-bracket destination and all three title forms accept `\<delim>`; the label now has its entry there too.

## Scope of the behavior change

Only lines whose label contains a backslash are affected, in two directions, both toward the spec:

- `[ref1\[\]]: url` is now a definition. That is the fix.
- `[ref\]: url` is no longer one. The `\]` is escaped, so the label never closes and the line is a paragraph — which means MD034 will now flag a bare URL there. It previously parsed as a definition with the label `ref\`. This is not separable from reading escapes at all: the rule that lets `\]` sit inside a label is the same rule that stops it from ending one.

Both were checked against a reference CommonMark implementation (markdown-it in `commonmark` mode) rather than against my reading of the spec: the first renders `[x][ref1\[\]]` as a resolved link, the second renders as a `<p>`.

`\` is ASCII punctuation, so `\\` is consumed as one escaped backslash and the alternation cannot be left mid-escape. A label ending on a lone `\` no longer matches, since `\\.` has nothing to consume — the paragraph reading again.

Deliberately not widened further, but there is something you should know, because this commit is what makes it reachable. There are three ref-def label patterns in the tree:

- `src/lint_context/link_parser.rs` — `(?:[^\]\\]|\\.)+` (this change)
- `src/rules/md052_reference_links_images.rs:18` — `(?:[^\[\]\\]|\\.|\[[^\]]*\])*`, escape-aware but a different shape
- `src/rules/md053_link_image_reference_definitions.rs:16` — `^\s*\[([^\]]+)\]:\s+(.+)$`, the exact pre-fix shape

MD053 reads escaped-label definitions from `ctx` but cannot match the same line with its own regex, so its multi-line-title continuation tracking (`:268`) skips them and its case-variant duplicate detection (`:514`) does too. Neither produces a wrong warning today — `fix_capability()` is `Unfixable` and `fix()` is a no-op — but if MD053 ever gains a real fix, the first one would delete such a definition and strand its title line as a paragraph. I left it alone: it is a separate, narrower bug that wants its own change. Say the word and I will file it, or take it in a follow-up.

The regression test covers the direction that *is* this PR's risk — a newly-visible definition being reported unused by MD053 — which holds because MD053 unescapes the definition and usage sides alike.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests added — `tests/regressions/md034_ref_def_escaped_label_issue_814_test.rs`, four cases: the reporter's document produces no MD034 warning; all three of its definitions reach `ctx.reference_definitions()` with their labels intact; `[ref\]:` is left a non-definition *and* its URL is reported bare, pinning the reversed direction; and a used escaped label is not reported unused by MD053. The first three fail on `main`; the fourth passes there and is a guard on this change's blast radius rather than a test of the bug.
- [x] All tests passing — full `cargo test` (`cargo nextest` was not available here; `make test-legacy` runs the same targets). The only failures are wall-clock assertions that trip on an unoptimized build — around ten of them, every one matching the `default-filter` your `ci` nextest profile already excludes as "flaky due to resource variability". The set is not even stable between two runs of the same tree (`md050` in one, `md051` in the next), and it fails the same way on a clean checkout with my commit stashed, which is the signature of timing noise rather than of this change. No correctness test fails.
- [x] Manual testing performed — the reporter's exact `test.md` and CLI invocation, failing before and clean after.

## Checklist

- [x] Code follows project style — `cargo fmt --check` clean; `cargo clippy --workspace --lib --bins --tests -- -D warnings -D clippy::uninlined_format_args` clean. The `lint-actions` half of `make lint` was not run, as no workflow files are touched.
- [x] Documentation updated (if needed) — none needed. MD034's documented behavior is unchanged; the parser it reads was wrong.

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/src/lint_context/link_parser.rs
+++ b/src/lint_context/link_parser.rs
@@ -69,6 +69,13 @@ fn extend_collapsed_byte_end(content: &str, link_type: LinkType, byte_end: usize
 // Mirrors the CommonMark §4.7 grammar closely enough for downstream consumers
 // (MD053, MD057, the MD054 round-trip pass) to see every valid ref def:
 //
+// - **Label** (group 1) ends at the first `]` that is not backslash-escaped
+//   (§6.3), so it permits `\]` — without that, `[ref\[\]]: url` is not read as
+//   a definition at all and its destination is left looking like a bare URL.
+//   `\` is excluded from the negated class so it has exactly one reading.
+//   §6.3 also forbids an *unescaped* `[` inside a label; that is still
+//   accepted here, as it was before.
+//
 // - **Destination** has two §6.6 forms: a bare sequence containing no spaces
 //   (group 3), or an angle-bracketed sequence that may contain spaces but no
 //   line endings or unescaped `<`/`>` (group 2). Without the angle form,
@@ -89,7 +96,7 @@ fn extend_collapsed_byte_end(content: &str, link_type: LinkType, byte_end: usize
 // out of `ctx.reference_defs` and breaking MD053/MD057 follow-ups.
 static REF_DEF_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(
-        r#"(?m)^[ ]{0,3}\[([^\]]+)\]:\s*(?:<((?:[^<>\n\\]|\\.)*)>|([^\s<][^\s]*))(?:\s+(?:"((?:[^"\\]|\\.)*)"|'((?:[^'\\]|\\.)*)'|\(((?:[^()\\]|\\.)*)\)))?$"#,
+        r#"(?m)^[ ]{0,3}\[((?:[^\]\\]|\\.)+)\]:\s*(?:<((?:[^<>\n\\]|\\.)*)>|([^\s<][^\s]*))(?:\s+(?:"((?:[^"\\]|\\.)*)"|'((?:[^'\\]|\\.)*)'|\(((?:[^()\\]|\\.)*)\)))?$"#,
     )
     .unwrap()
 });
--- a/tests/regressions/md034_ref_def_escaped_label_issue_814_test.rs
+++ b/tests/regressions/md034_ref_def_escaped_label_issue_814_test.rs
@@ -1,0 +1,96 @@
+//! Tests for reference definitions whose link label contains an escaped
+//! closing bracket (Issue #814)
+//!
+//! CommonMark §6.3: a link label ends at the first `]` that is **not**
+//! backslash-escaped. `[ref\[\]]: url` is therefore a reference definition
+//! with the label `ref\[\]`, but the label group of `REF_DEF_PATTERN` did not
+//! read escapes, so the line was not recognised as a definition at all and
+//! MD034 saw its destination as a bare URL.
+
+use rumdl_lib::config::MarkdownFlavor;
+use rumdl_lib::lint_context::LintContext;
+use rumdl_lib::rule::Rule;
+use rumdl_lib::rules::{MD034NoBareUrls, MD053LinkImageReferenceDefinitions};
+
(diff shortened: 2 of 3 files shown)


rvben:
Thanks for this. `main` already carries the same `REF_DEF_PATTERN` change in
f024e4df (2026-08-13), plus the MD053 regex this PR leaves alone and a test
for the reversed direction (`test_unterminated_label_is_not_a_reference_definition`),
which your regression file also pins. Nothing here is missing from `main`, so I
am closing this rather than asking you to rebase onto a change that already
includes it. #814 (with @gifnksm's reproduction) is closed with v0.2.56, which
ships the fix.

PR #826, opened 2026-08-21
docs: config.rs became a module directory

melbinjp:
The JSON Schema Generation section tells a contributor to regenerate the schema after modifying `src/config.rs`. That file was extracted into a module directory in `3cb9af37` ("refactor(config): extract into module directory with 8 submodules", 2026-02-06), and `src/config.rs` no longer exists on `main`.

Changed to `src/config/`, since the instruction is about the configuration structures rather than about one file, and they are now spread across the submodules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1357,7 +1357,7 @@ make test
 
 ### JSON Schema Generation
 
-If you modify the configuration structures in `src/config.rs`, regenerate the JSON schema:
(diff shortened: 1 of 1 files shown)


rvben:
Thanks, @melbinjp—good catch. `src/config/` is the correct reference after the module refactor. Merging this now.

PR #827, opened 2026-08-26
feat(flavor): add support for Markdown with Gherkin (MDG)

daiksud:
## Description

Adds first-class support for [Markdown with Gherkin](https://github.com/cucumber/gherkin/blob/main/MARKDOWN_WITH_GHERKIN.md) as the `mdg` flavor (alias `markdown_with_gherkin`). Files ending in `.feature.md` are auto-detected, while ordinary Markdown behavior remains unchanged. The flavor protects Gherkin-sensitive syntax during linting and formatting but does not validate Gherkin.

## Type of Change

- [ ] Bug fix (non-breaking change fixing an issue)
- [x] New feature (non-breaking change adding functionality)
- [ ] Documentation update

## Testing

- [x] Tests added/updated

<details>
<summary>Manual testing details</summary>

Validated 16 representative `.feature.md` files with `@cucumber/gherkin` v42 by comparing compiled pickles before and after `rumdl fmt --flavor mdg`, normalizing only generated IDs and source locations.

- MDG preserved every pickle set, produced idempotent output, and passed `fmt --check`; malformed input remained rejected.
- Verified suffix auto-detection, explicit flavor aliases and overrides, and `[per-file-flavor]` mappings.
- Verified report-only behavior for MD013, MD034, and MD040, and one warning per process for incompatible MD003, MD026, MD046, MD048, and MD055 overrides.
- Reproduced the documented limitations, control cases, and intentionally unchanged rule behavior.
- Standard flavor changed 11 pickle sets on the same corpus; MDG changed none.

Testing used an external scratch workspace; no fixtures or generated artifacts were added.

</details>

## Checklist

- [x] Code follows project style (`make fmt` and `make lint`)
- [x] Documentation updated (if needed)
- [x] No unrelated changes included

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,7 @@ rumdl supports multiple Markdown flavors to accommodate different documentation 
 | [kramdown](docs/flavors/kramdown.md)         | Jekyll / kramdown            | IALs, ALDs, extension blocks                        |
 | [azure_devops](docs/flavors/azure_devops.md) | Azure DevOps Wiki            | Colon code fences (:::lang ... :::)                 |
 | [myst](docs/flavors/myst.md)                 | MyST / Jupyter Book / Sphinx | Directives, roles, `%` comments                     |
+| [mdg](docs/flavors/mdg.md)                   | Markdown with Gherkin        | Gherkin-safe headings, tags, Doc Strings, tables    |
 
 ### Configuring Flavors
 
@@ -628,7 +629,8 @@ Or configure per-file patterns:
 "**/*.qmd" = "quarto"
 ```
 
-When no flavor is configured, rumdl auto-detects based on file extension (`.mdx` → mdx, `.qmd`/`.Rmd` → quarto, `.md` → standard).
+When no flavor is configured, rumdl auto-detects from the file name: `.mdx` → mdx, `.qmd`/`.Rmd` → quarto, `.md` → standard.
+A name ending in `.feature.md` matches that compound suffix before the plain `.md` rule, so those files are linted as mdg.
 
 For complete flavor documentation, see the [Flavors Guide](docs/flavors.md).
 
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -54,17 +54,18 @@ Most tools treat Markdown as a single dialect and rely on configuration or plugi
 
 **rumdl** has built-in flavor support that adjusts rule behavior for specific documentation systems:
 
-| Flavor       | Target system       | Example adjustments                                                          |
-| ------------ | ------------------- | ---------------------------------------------------------------------------- |
-| standard     | CommonMark + GFM    | Baseline behavior (GFM extensions included by default)                       |
-| mkdocs       | MkDocs / Material   | Admonitions, tabs, mkdocstrings                                              |
-| mdx          | MDX                 | JSX components, ESM imports                                                  |
-| obsidian     | Obsidian            | Callouts, wikilinks, Dataview                                                |
-| pandoc       | Pandoc Markdown     | Fenced divs, attribute lists, citations, definition lists, math, grid tables |
-| quarto       | Quarto / RMarkdown  | Citations, shortcodes, executable blocks                                     |
-| kramdown     | Jekyll / kramdown   | Attribute lists, TOC markers                                                 |
-| azure_devops | Azure DevOps wikis  | Colon code fences (`:::mermaid … :::`) treated as opaque code blocks         |
-| myst         | MyST / Jupyter Book | Directives (`:::{name}`), roles (`` {role}`text` ``), `%` comments           |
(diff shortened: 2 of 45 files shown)


daiksud:
Hi @rvben, I’d like to add support for [Markdown with Gherkin](https://github.com/cucumber/gherkin/blob/main/MARKDOWN_WITH_GHERKIN.md) as a rumdl flavor. Could you review this PR? Thank you!

rvben (commented):
Really nice work. The research here is better than the spec deserves: you caught
that `GherkinInMarkdownTokenMatcher` only accepts backtick Doc String fences
even though the prose says "GFM fenced code blocks", and that the table indent
counts tabs. Matching a colon instead of a keyword table is the right call too.

I built the branch and ran cucumber's own `testdata/good/*.feature.md` through
`rumdl fmt`, parsing before and after with `@cucumber/gherkin`. Three of the
five come out with an identical AST and fmt is idempotent on all five. The two
that change, plus one inconsistency:

**MD040 changes the AST.** On `docstrings.feature.md`, fmt labels the unlabelled
Doc String fence `text`, and the parser goes from `mediaType: null` to
`mediaType: "text"`. That reaches step definitions, so it would stamp a media
type on every unlabelled Doc String the first time someone runs fmt. The docs
list MD040 as deliberately left unchanged. Could it take MD013's shape (report,
no fix), or at least move to Limitations?

**MD003 rewrites a setext heading on `misc.feature.md`**, which changes the
parsed feature name to one with a literal `# ` on the front. Not a regression,
MD003 does that under standard too, but MDG removes the `style = "setext"`
escape hatch: an explicit setext config is overridden with no notice, while
MD046, MD048 and MD055 all warn for the same class of override. MD003 is also
the one whose fix rewrites headings on disk. Intentional?

**MD026's warning is behind `should_skip`.** The skip at
md026_no_trailing_punctuation.rs:233 tests the post-override set, so the notice
only fires if the file happens to contain one of the *other* punctuation
characters. A `.feature.md` with `#### Examples:` and nothing else stays silent;
add one unrelated "Some prose." and the warning appears.

Two questions: why is MD034 disabled outright rather than report-without-fix
like MD013? And `is_tag_line` is stricter than the reference regex, so
`` `@comment_tag1` #a comment `` in `tags.feature.md` is not recognised.
Harmless as far as I can tell (the AST is unchanged either way), just checking
it is deliberate.

Small one: the check help now says "or myst (...), or mdg (...)".

The suite passes locally (12795 tests). CI has not run since you are a
first-time contributor, I will approve the workflows.

daiksud:
:robot: Thank you for the careful review and for validating the output with Cucumber's parser. I used the feedback to make the MDG policy more consistent. The MDG-specific behavior now falls into three categories:

1. **Report, but do not auto-fix when there is no generally safe, semantics-preserving rewrite:** MD013, MD034, and MD040.
2. **Enforce the form required by Gherkin and emit a one-time warning when an incompatible explicit setting is overridden:** MD003, MD026, MD046, MD048, and MD055.
3. **Other syntax-aware compatibility handling:** MD022, MD060, and MD063.

For the specific review points:

- **MD040** now follows MD013's report-only pattern under MDG. It reports missing or inconsistent Doc String media types but provides no fix, so formatting cannot change `mediaType: null` to `"text"` or normalize an explicitly selected media type. Standard behavior is unchanged.
- **MD034** also reports without fixing under MDG. Angle brackets are Gherkin placeholder syntax, while the appropriate alternative—an explicit Markdown link or disabling MD034—depends on context. The diagnostic now explains that choice. I also fixed the category prefilter so email/XMPP-only documents are checked.
- **MD003's plain ATX conversion is intentional**, since that is the heading form Gherkin recognizes. It now behaves consistently with MD046 and the other enforced-setting rules: an incompatible explicit fixed style is overridden with a one-time warning. Explicit `atx` and `consistent` remain silent. The known Setext AST change is documented.
- **MD026's warning now occurs before `should_skip`**, so a document containing only `#### Examples:` still reports the override. The notice is process-wide, including runs with multiple per-directory configuration groups.
- The stricter **tag-line matcher was not intentional**. It now follows Cucumber's `` /`(@[^`]+)`/g `` matcher, including the comment case from `tags.feature.md`.
- The duplicated conjunction in the help text has been fixed.

The related flavor and rule documentation has also been updated.

Locally, all 12,801 tests pass, along with formatting, Clippy, and documentation linting. I also reran Cucumber's five `testdata/good/*.feature.md` fixtures: formatting is idempotent for all five, and four retain an identical AST. The remaining difference is the documented MD003 Setext case in `misc.feature.md`.

Addressed in 864cc8fd. Thanks again for the thorough parser-level review.

daiksud:
Thank you!

rvben:
Thanks @daiksud, for your work! Merged your PR and released a new version with the new flavor. 💪

Issue #704, opened 2026-06-30
Links may get broken up since `v0.2.25`


denialhaag:
### rumdl version

0.2.25 and 0.2.26

### Installation method

cargo install

### Bug description

Links may get broken up since `v0.2.25` if the `reflow-mode` is set to `normalize`.

### Steps to reproduce

Run `rumdl check --fix` on this:

```markdown
I have set `reflow-mode` to `normalize`, and [this is a lot of text for a simple link to github.com][github].

[github]: https://github.com/
```

The result is this:

```markdown
I have set `reflow-mode` to `normalize`, and [this is a lot of text for a simple
link to github.com][github].

[github]: https://github.com/
```

### Expected behavior

The link does not get broken up.

### Actual behavior

The link does get broken up.

### Configuration

```toml
[MD013]
line-length = 80
reflow = true
reflow-mode = "normalize"
```

### Operating system

macOS

rvben:
Thanks @denialhaag, fixed in **v0.2.27**: reflow keeps reference-style links atomic again, so the link text no longer wraps mid-link.

Issue #412, opened 2026-02-11
Bug: `MD013` `semantic-line-breaks` reflow splits link labels across lines


pygarap:
## Summary

With `MD013` configured for reflow using `reflow-mode = "semantic-line-breaks"`, `rumdl check --fix` can insert a newline inside the link label (`[...]`), splitting it across lines.

This hurts readability and can produce surprising diffs, especially for reference-style sentences that should stay intact.

## Reproduction

Config (minimal):
```toml
[MD013]
line-length = 60
reflow = true
reflow-mode = "semantic-line-breaks"
```

Input:
```markdown
Reference: [Widget Entrypoint - Location](https://example.com/docs/widget-entrypoint-location).
```

Run:
```sh
rumdl check --fix path/to/file.md
```

Actual output:
```markdown
Reference: [Widget Entrypoint -
Location](https://example.com/docs/widget-entrypoint-location).
```

Expected output:
```markdown
Reference: [Widget Entrypoint - Location](https://example.com/docs/widget-entrypoint-location).
```

## Expected behavior

Reflow should treat Markdown link constructs as atomic and avoid inserting line breaks inside:

- link labels: `[... ]`
- link destinations: `( ... )`

If wrapping is needed, wrap outside the link construct (before the link or after it), but never inside `[...]` or `(...)`.

## Notes

- This is most visible when the label contains spaced punctuation like ` - `, where the reflow logic may choose that as a break point.
- A regression test should ensure the fix is idempotent under `rumdl check --fix`.

## References

- https://github.com/rvben/rumdl/blob/main/docs/RULES.md
- https://github.com/rvben/rumdl/blob/main/docs/global-settings.md

rvben:
Fixed in v0.1.19. The reflow engine now tracks element boundaries when searching for line-break positions. Markdown links (`[text](url)`) are treated as atomic units and are never split across lines — line breaks are inserted before or after the link construct instead.

Thanks for reporting this, @pygarap.

PR #828, opened 2026-08-27
fix(MD013): support line wrapping within markdown link text

chandlerc:


rvben:
Thanks for the PR, @chandlerc. I'd say yes to breaking inside link text, I want this in. The one thing I need changed is that it lands as an option rather than a new default.

#412, #704 and #252 all asked for links to stay atomic. I built this branch and #412 and #704 come back with the output verbatim from their reports, so I think it's best to keep the current default. That only constrains the default though. Add `MD013.reflow-break-link-text` (name negotiable), default false, and you set it once in your config and get exactly what you built here.

For the routing you already have the mechanism: links are `is_hard` from the match you added in #779 (`text_reflow.rs:3452`), and `is_inside_element_filtered` at `:3364` honours that plus `atomic_spans`. The new branch at `:3648` consults neither, so there are two notions of atomic in the file now. Making links soft under the new option is the smaller change and keeps one definition, same shape as `emphasis_spans` in #736.

Three things that will bite regardless of which way the default goes:

- `fmt` can emit a file that `check` then rejects and cannot fix. `is_url_exempt` (`:3689`) requires `suffix.ends_with(')')`, so a titled link loses the URL exemption once split. Your own fixture at `md013_reflow_link_text_breaking_test.rs:165` at `line-length = 45` does it: `check` reports `69 exceeds 45`, no `[*]` marker, and a second `fmt` is a no-op.
- `helpers.rs:454` uses `inner.len()`, so that exemption measures bytes and ignores `length_mode`. Two link texts of the same display width behave differently depending on encoding.
- Full and collapsed references break, but shortcut references and reference images stay atomic (`text_reflow_test.rs:279`, `:287`, `:327`). The four forms should agree either way.

The `+2` extension for collapsed reference spans in `all_link_spans` is a nice fix independent of all of the above, and I would take that on its own.

PR #836, opened 2026-08-31
feat(md063): add consistent style option

### LINKED ISSUE #833: Suggestion: Add `consistent` support to `style` option in MD063
### Problem or motivation

Creating a user-level configuration that is as generally useful as possible is difficult in the context of MD063 because a specific style must be chosen instead of just enabling it to check style consistency.

### Proposed solution

I think adding support for `consistent` to the `style` option in MD063 would make it more generally useful. It would work similarly to MD003 where different styles are possible to specify, but `consistent` just checks that the style is consistent.

### Alternatives considered

The alternative for keeping things general for user-level is to keep the rule disabled, which feels unfortunate.

### Feature category

Improve existing rule

### Example markdown

```markdown
<!-- document 1 -->
# Title Case Heading

## Another Title Case Heading



<!-- document 2 -->
# Sentence case heading

## Another sentence case heading


...would both be valid with:


[MD063]
style = "consistent"
```

### Contribution

### Additional context

The envisioned functionality would be similar to MD003.

DavidPandleton:
Closes #833.

## Problem

MD063's `style` option only accepts a fixed style (`title-case`, `sentence-case`, `all-caps`), so a user-level config can only enable the rule for documents written in one specific style. There was no way to just "check that the style is consistent" regardless of which style the document uses, forcing users who want consistency to keep the rule disabled.

## Fix

Added a `consistent` style value that infers the target style from the first checkable heading and enforces it throughout the document:

```toml
[MD063]
style = "consistent"
```

Both of these would now pass:
```markdown
# Title Case Heading
## Another Title Case Heading
```

```markdown
# Sentence case heading
## Another sentence case heading
```

But a document that mixes styles warns on the headings that deviate from the first heading's style, and `--fix` normalizes them to it.

## Implementation

- `HeadingCapStyle::Consistent` variant with serialization/deserialization support in `md063_config.rs`
- `get_effective_style(ctx)` resolves the target style once per document (mirrors MD003's approach): in consistent mode it scans for the first heading the rule would actually check (respecting level filter, validity, and code-block exclusion) and infers its style
- `detect_heading_cap_style(text, flavor)` tries each concrete style by applying it and comparing against the original; TitleCase is tried first as the deterministic tiebreaker for text valid under multiple styles
- `apply_capitalization` now takes an explicit `style` parameter so check() and fix() can pass the resolved style; the two internal matches are updated and `unreachable!()` guards prevent Consistent from reaching the per-segment dispatch
- Warning messages name the resolved style (e.g. "Heading should use title case"), not the literal "consistent"
- Docs updated with the new option value

## Testing

Added 8 rule tests covering: all-title-case docs, all-sentence-case docs, all-caps docs, mixed styles warning, sentence-first docs enforcing sentence case, fix normalizing to the first style, resolved style name in messages, and code-block headings being ignored for detection. Plus config serialization/deserialization tests for `consistent`.

All existing MD063 tests (136) still pass; full lib suite passes (2 pre-existing `text_reflow` perf-test failures unrelated to this change).

### DIFF
--- a/docs/md063.md
+++ b/docs/md063.md
@@ -4,7 +4,7 @@ Aliases: `heading-capitalization`
 
 ## What This Rule Does
 
-Enforces consistent capitalization styles for markdown headings. This rule supports title case, sentence case, and all caps styles.
+Enforces consistent capitalization styles for markdown headings. This rule supports title case, sentence case, all caps, and consistent styles.
 
 **Note:** This rule is disabled by default. You must explicitly enable it in your configuration.
 
@@ -71,7 +71,7 @@ Enforces consistent capitalization styles for markdown headings. This rule suppo
 
 | Option                        | Type     | Default        | Description                                                        |
 | ----------------------------- | -------- | -------------- | ------------------------------------------------------------------ |
-| `style`                       | string   | `"title-case"` | Capitalization style: `title-case`, `sentence-case`, or `all-caps` |
+| `style`                       | string   | `"title-case"` | Capitalization style: `title-case`, `sentence-case`, `all-caps`, or `consistent` |
 | `lowercase-words`             | string[] | *(see below)*  | Words to keep lowercase in title case                              |
 | `ignore-words`                | string[] | `[]`           | Words to preserve exactly (brand names)                            |
 | `preserve-cased-words`        | boolean  | `true`         | Auto-preserve words with internal capitals                         |
--- a/src/rules/md063_heading_capitalization.rs
+++ b/src/rules/md063_heading_capitalization.rs
@@ -729,7 +729,12 @@ impl MD063HeadingCapitalization {
     }
 
     /// Apply capitalization to heading text
-    fn apply_capitalization(&self, text: &str, flavor: crate::config::MarkdownFlavor) -> String {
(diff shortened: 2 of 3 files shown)


rvben (changes requested):
Thanks for adding the opt-in `consistent` mode. The approach is well scoped, and the tests cover the main title-case and sentence-case paths. I found two issues to address before merging:

1. Applied to current `main`, this PR does not compile. The newer multi-line Setext heading path (`fix_setext_heading_span`) still calls `apply_capitalization` with two arguments, but this PR changes it to require a style argument. Please update that path and its callers when rebasing.

2. Style detection misses an ALL CAPS mismatch with the default configuration. For `# HELLO WORLD` followed by `## Goodbye Moon`, `check()` returns no warnings. The detector tries title case first, and acronym preservation lets it accept the first heading. Please make this case detect ALL CAPS and add a regression test.

One behavior worth documenting as you update this: a short first heading such as `# API` fits multiple styles. Please state how `consistent` resolves that ambiguity so users know what later headings will be checked against.

I reproduced both blocking issues locally. Happy to review again after the update.

PR #837, opened 2026-08-31
feat(cli): add --no-code-block-tools and --only-code-block-tools flags

### LINKED ISSUE #829: Add `--no-code-block-tools` and `--only-code-block-tools` CLI flags
## Summary

Add two complementary CLI flags that control `code-block-tools` for a single invocation without discarding unrelated rumdl configuration.

## Proposed behavior

- `--no-code-block-tools`: Disable `code-block-tools` for the current invocation while preserving the configured Markdown rules and other loaded settings.
- `--only-code-block-tools`: Enable `code-block-tools` for the current invocation, preserve its configured languages and tools, and disable all regular Markdown rules.
- Support both flags with `rumdl check`, `rumdl check --fix`, and `rumdl fmt`.
- Make the flags mutually exclusive.
- Give either explicit CLI flag final precedence over configuration files and any environment variable that can affect the same behavior.

## Motivation

`--no-config` disables `code-block-tools` only by discarding every loaded configuration value, including unrelated rule and file-selection settings. `--disable all` can approximate a `code-block-tools`-only run, but only when `code-block-tools` is already enabled and configured.

Dedicated flags would make both modes explicit, symmetric, and predictable.

## Usage

Given this configuration:

```toml
[code-block-tools]
enabled = true

[code-block-tools.languages]
python = { lint = ["ruff:check"], format = ["ruff:format"] }
```

Run regular Markdown checks without invoking configured code-block tools:

```console
$ rumdl check --no-code-block-tools docs/
```

Run only the configured code-block linters:

```console
$ rumdl check --only-code-block-tools docs/
```

Run only the configured code-block formatters:

```console
$ rumdl fmt --only-code-block-tools docs/
```

## References

- #331
- [Code Block Tools](https://github.com/rvben/rumdl/blob/main/docs/code-block-tools.md)
- [CLI Commands](https://github.com/rvben/rumdl/blob/main/docs/usage/cli.md)
- [Rules Reference](https://github.com/rvben/rumdl/blob/main/docs/rules.md)

DavidPandleton:
## Summary

Adds two complementary CLI flags that control code-block-tools for a single invocation **without discarding unrelated rumdl configuration**:

- `--no-code-block-tools` — disables code-block-tools for this run while Markdown rules and all other settings stay active
- `--only-code-block-tools` — enables code-block-tools (preserving its configured languages/tools) and disables all regular Markdown rules

Both flags are mutually exclusive (clap `conflicts_with`) and take final precedence over config files and environment variables. They are available on `rumdl check`, `check --fix`, and `rumdl fmt` (added to `SharedCliArgs`, so every command that flattens it gets them, including watch mode and stdin paths).

## Why

Closes #829. Users on shared machines or CI want to toggle the (potentially slow/external) code-block-tools per invocation without editing `.rumdl.toml`:

```bash
# CI: only run the fast Markdown rules
rumdl check --no-code-block-tools .

# Focused run: only the code-block tools
rumdl check --only-code-block-tools .
```

## Implementation

- `src/cli_types.rs` — both flags on `SharedCliArgs`, mutually exclusive
- `src/cli_utils.rs` — `apply_cli_overrides` sets `code_block_tools.enabled` at `ConfigSource::Cli` precedence
- `src/resolution.rs` — `apply_cli_config_overrides` mirrors the override onto subdirectory configs so the flag wins everywhere
- `src/file_processor/discovery.rs` — `--only-code-block-tools` short-circuits `get_enabled_rules_from_checkargs` to no rules
- `tests/cli/cli_cbt_flags_test.rs` — 5 integration tests using the built-in `rumdl` tool for embedded markdown blocks (no external binary needed in CI)

## Test plan

- `cargo test --test lib -- cli_cbt_flags_test` — 5/5 pass
- `cargo test --lib` — 7184 pass, 2 pre-existing perf-test failures (`text_reflow`, unrelated)
- `make fmt-check`, `cargo clippy --lib` — clean

### DIFF
--- a/src/cli_types.rs
+++ b/src/cli_types.rs
@@ -115,6 +115,25 @@ pub struct SharedCliArgs {
     /// default.
     #[arg(long, help = "Treat configuration warnings as errors (exit code 2)")]
     pub deny_config_warnings: bool,
+
+    /// Disable code-block-tools for this invocation while preserving the
+    /// configured Markdown rules and other loaded settings.
+    #[arg(
+        long,
+        conflicts_with = "only_code_block_tools",
+        help = "Disable code-block-tools for this invocation (Markdown rules still run)"
+    )]
+    pub no_code_block_tools: bool,
+
+    /// Run only code-block-tools: enable them for this invocation, preserve
+    /// their configured languages and tools, and disable all regular Markdown
+    /// rules.
+    #[arg(
+        long,
+        conflicts_with = "no_code_block_tools",
+        help = "Run only code-block-tools (disables all regular Markdown rules)"
+    )]
+    pub only_code_block_tools: bool,
 }
 
 #[derive(Args, Debug)]
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -45,6 +45,15 @@ pub fn apply_cli_overrides(sourced: &mut rumdl_config::SourcedConfig, args: &Che
             .collect();
         sourced.global.unfixable = rumdl_config::SourcedValue::new(rules, rumdl_config::ConfigSource::Cli);
     }
+
+    // Apply --no-code-block-tools / --only-code-block-tools override.
+    // Either explicit CLI flag wins over configuration files and any
+    // environment variable that can affect the same behavior.
+    if args.no_code_block_tools || args.only_code_block_tools {
+        let mut cbt = sourced.code_block_tools.value.clone();
+        cbt.enabled = args.only_code_block_tools;
+        sourced.code_block_tools = rumdl_config::SourcedValue::new(cbt, rumdl_config::ConfigSource::Cli);
+    }
 }
 
 /// Resolve the lint output format with the standard precedence:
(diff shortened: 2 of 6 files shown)


rvben:
Thanks for the PR, and sorry for going quiet on it. I ended up fixing #829
differently on main (`fc410f77`, in v0.2.65) and should have said so here
instead of leaving you hanging.

Closing this as superseded. The `--no-code-block-tools` half is fine, but
`--only-code-block-tools` falls apart once the built-in `rumdl` tool is
configured for `markdown` blocks: the early `return Vec::new()` in
`get_enabled_rules_from_checkargs` also empties the rule set that
`embedded_lint` uses for fenced Markdown, so those findings disappear, `fmt`
stops fixing them, and `--stdin` reports a clean run on a path where the tools
never execute at all. Easy to miss, since `test_only_code_block_tools_flag`
asserts "No issues found" for exactly that setup.

What shipped keeps two rule sets (outer document vs fenced Markdown), rejects
the flags for stdin, and warns when only mode runs with no tool configured.
Several of your choices matched, the subdirectory config mirroring in
particular.

Appreciated, and happy to see more from you.

PR #855, opened 2026-09-05
fix(MD051): delete the emoji marker pass from the GitHub anchor style

### LINKED ISSUE #854: MD051: `§` survives the GitHub slug, and an emoji not surrounded by spaces gets one hyphen too many — both from the emoji marker pass
### rumdl version

0.2.65. Reproduces identically on 0.2.35 and 0.2.63.

### Installation method

cargo install (also reproduced through `rumdl_lib::lint` as a library consumer)

### Bug description

The GitHub anchor style has an emoji marker pass (`process_emoji_sequences` in `src/utils/anchor_styles/github.rs`) that replaces each emoji with a `§EMOJI§` marker before the character pass, then rewrites the markers into hyphens at the end. It causes two divergences from GitHub.

**1. A section sign (`§`, U+00A7) in a heading survives the slug.** `§` is the marker's delimiter, and the character pass preserves it so the marker survives to the reassembly step. A `§` written in the heading is indistinguishable from one the pass inserted, so it survives too:

```rust
// src/utils/anchor_styles/github.rs:275 (unchanged from v0.2.35 to v0.2.65)
} else if c == '§' {
    // Preserve our marker character
    result.push(c);
}
```

GitHub strips the character, so for `## A §1 B` MD051 rejects `#a-1-b` (the anchor a browser resolves) and accepts `#a-§1-b` (an anchor no renderer produces). `§` is common in specification prose (`## 9. Worked Examples — the ridl §13 Contracts`), where every correct link to such a heading is reported.

**2. An emoji not surrounded by spaces gets one hyphen too many.** GitHub's anchor filter keeps `[\p{Word}\- ]`, drops every other character and turns each space into a hyphen, so an emoji is deleted and the spaces around it stay. The marker pass instead pops the spaces around the emoji and re-adds one or two hyphens by position, which is exact only when the emoji is surrounded by spaces:

| Heading | GitHub (github-slugger 2.0.0) | rumdl 0.2.65 |
| --- | --- | --- |
| `A 🚀 B` | `a--b` | `a--b` |
| `A🚀 B` | `a-b` | `a--b` |
| `A 🚀B` | `a-b` | `a--b` |
| `A🚀B` | `ab` | `a--b` |
| `A °🚀 B` | `a--b` | `a---b` |
| `Notes… 🚀 done` | `notes--done` | `notes---done` |
| `Notes… done` | `notes-done` | `notes-done` |

The last two rows show a related effect: the pass treats every `is_emoji_or_symbol` character as a symbol, a range that includes General Punctuation (`…`, `•`, `“`), but the pass only runs when the heading also contains an emoji from its quick-check ranges. So the slug of `…` depends on whether an emoji appears elsewhere in the same heading.

### Steps to reproduce

1. Create `repro.md`:

   ```markdown
   # T

   [l](#a-1-b)

   ## A §1 B

   x
   ```

2. Run `rumdl check repro.md`.

3. For the mirror, change the link to `[l](#a-§1-b)` and run again. This anchor resolves in no renderer.

4. For the emoji case, use `[l](#a-b)` and `## A🚀 B`.

### Expected behavior

`rumdl check repro.md` reports nothing: `#a-1-b` is the anchor GitHub produces for `## A §1 B`. The mirror, `[l](#a-§1-b)`, is reported. `#a-b` resolves `## A🚀 B`.

### Actual behavior

```console
$ rumdl check --no-config repro.md
repro.md:3:1: [MD051] Link anchor '#a-1-b' does not exist in document headings

Issues: Found 1 issues in 1 file (10ms)

$ rumdl check --no-config mirror.md

Success: No issues found in 1 file (7ms)
```

`heading_to_fragment("A §1 B")` returns `"a-§1-b"`; `heading_to_fragment("A🚀 B")` returns `"a--b"`.

### Configuration

None. Defaults, `flavor` unset.

### Operating system

macOS

### Additional context

**`§` is the only character that survives.** Substituting each of 73 characters GitHub strips (ASCII punctuation except `-` and `_`; Latin-1 U+00A1–U+00BF except the letters, numbers and soft hyphen; `×` `÷`; `– — ‘ ’ “ ” ‹ › † ‡ • … ‰ € ™ ≈ ≠ ≤ ≥ ∞`) into `## A X1 B` with the link left at `#a-1-b`, MD051 fires for exactly one of them: `§`. Every other symbol, including the others in the same Unicode category (`°`, `©`, `×`, `¶`), is stripped the way GitHub strips it. Same result on 0.2.35, 0.2.63 and 0.2.65.

**The pass is redundant with the character pass.** With `process_emoji_sequences` not called at all, every existing test in `github.rs` (33, including the emoji, flag, ZWJ-family and keycap cases) passes unchanged, and all the rows above match GitHub. ZWJ sequences, flags and skin-tone modifiers come out the same because each component is deleted and no space sits between them.

`kramdown_gfm.rs` uses the same delimiter for its own markers but is not affected: it preserves `§` only when it heads one of its marker strings, and a bare `§` falls through to the normal branch and is dropped.

A PR follows (linked below) that deletes the pass. The smaller alternative — keep the pass and drop the heading's own `§` before it runs — fixes the section sign only; it does not touch the hyphen count, since that comes from the pass itself. I have that variant ready as well if a one-line change is preferred for a first step.

Found through [prim](https://github.com/driftsys/prim), which runs MD051 as part of its selected rule set and carries a local workaround for the `§` case ([driftsys/prim#180](https://github.com/driftsys/prim/issues/180)) until it is fixed here.

stasson:
## Description

Fixes #854.

GitHub's anchor filter keeps `[\p{Word}\- ]`, drops every other character and turns each space into a hyphen, so an emoji is deleted and the spaces around it stay. The character pass in `heading_to_fragment_internal` already does exactly that. The emoji marker pass in front of it (`process_emoji_sequences`) replaced each emoji with a `§EMOJI§` marker, popped the spaces around it, and re-added one or two hyphens by position at the end — a reconstruction that is exact only when the emoji is surrounded by spaces, and whose delimiter was in band, so a section sign written in the heading survived the slug.

This PR deletes the pass, the step-11 marker rewrite, the `§` branch of the character pass, and the two helpers only the pass used (`is_emoji_or_symbol` and `is_regional_indicator` in this file; `is_keycap_base` stays, the character pass uses it; `common::is_emoji_or_symbol` is untouched). An emoji now falls through the character pass like any other symbol. Nothing `§`-specific is left in the file.

### What changes for a user

All toward GitHub's output, verified against github-slugger 2.0.0:

| Heading | before | after |
| --- | --- | --- |
| `A §1 B` | `a-§1-b` | `a-1-b` |
| `A🚀 B`, `A 🚀B` | `a--b` | `a-b` |
| `A🚀B` | `a--b` | `ab` |
| `A °🚀 B` | `a---b` | `a--b` |
| `Notes… 🚀 done` | `notes---done` | `notes--done` |
| `Top 3☀️ days` | `top-3--days` | `top-3️-days` (as `⭐️` already behaved) |

Consequences downstream of the slug: MD051 stops reporting correct links to a `§` heading and starts reporting a link that was written to the old slug (`#a--b` for `## A🚀 B`), which GitHub never resolved; MD073's auto-fix regenerates a marker-based TOC with the corrected anchors, so a TOC rumdl wrote for such a heading changes on the next `--fix`; MD080 and the LSP rename compute the same slug and move with it.

### What does not change — tested

ZWJ family sequences (the pass ran before `sanitize_unicode` strips U+200D; each component is deleted and no space sits between them), flags, keycaps, skin-tone modifiers, text-style emoji with VS16, symbols outside the pass's quick-check ranges (`→`, `∑`, `⭐`), and every security filter. The other anchor styles are not touched.

### The smaller alternative, and why I did not take it

Keeping the pass and dropping the heading's own `§` before it runs is a one-line fix for the section sign, and I have it ready on a branch. It does not touch the hyphen count, since that comes from the pass itself, so the emoji rows above stay as they are. If you would rather land the one-liner first and consider the deletion separately, say so and I will swap the branch.

The one thing I cannot measure: the module says its cases were verified against GitHub Gists, and a verified emoji case that never became a test would be invisible to me. All 33 existing tests in the module pass without the pass, and GitHub's published filter has no emoji special case at all.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests added/updated — `test_github_section_sign_is_stripped` (ten headings) and `test_github_emoji_keeps_the_spaces_around_it` (eleven), expected values from github-slugger 2.0.0. The first fails on `main` with `"a-§1-b"` vs `"a-1-b"`, the second with `"a--b"` vs `"a-b"`.
- [x] All tests passing (`make test-dev`) — 13,140 run, 13,140 passed; `cargo test --doc` 21 passed.
- [x] Manual testing performed — `rumdl check` on the reproductions in #854: the `§` and emoji files report nothing, the mirror `[l](#a-§1-b)` is reported.

## Checklist

- [x] Documentation updated (the module doc no longer describes the pass)

### DIFF
--- a/src/utils/anchor_styles/github.rs
+++ b/src/utils/anchor_styles/github.rs
@@ -13,7 +13,7 @@
 //! 6. Multi-character pattern replacement (-->, <->, ==>, ->)
 //! 7. Special symbol replacement (& → --, © → --)
 //! 8. Character processing (preserve letters, digits, underscores, hyphens)
-//! 9. Space → single hyphen, emojis → single hyphen
+//! 9. Space → single hyphen; emojis and other symbols are deleted, keeping the spaces around them
 //! 10. No leading/trailing trimming (unlike kramdown)
 //!
 //! Security measures implemented:
@@ -23,7 +23,6 @@
 //! - Zero-width character stripping
 //! - Control character filtering
 //! - ReDoS-resistant regex patterns with complexity limits
-//! - Comprehensive emoji detection including country flags and keycaps
 
 use regex::Regex;
 use std::sync::LazyLock;
@@ -78,7 +77,6 @@ static HTML_COMMENT_PATTERN: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"<!-
 /// - Zero-width character removal
 /// - Control character filtering
 /// - ReDoS-resistant regex patterns
-/// - Comprehensive emoji detection
 ///
 /// # Examples
 /// ```
@@ -121,23 +119,13 @@ fn heading_to_fragment_internal(heading: &str) -> String {
     // NFC normalization ensures canonical representation
     let normalized: String = heading.nfc().collect();
 
-    // Step 3: Handle emoji sequences BEFORE sanitizing ZWJ
-    // This preserves multi-component emojis and keycaps
-    // Quick optimization: skip if clearly no emojis (common case)
-    let emoji_processed = if normalized.chars().any(|c| {
-        let code = c as u32;
-        // Quick check for common emoji ranges
-        (0x1F300..=0x1F9FF).contains(&code) || // Most emojis
-        (0x2600..=0x26FF).contains(&code) ||   // Misc symbols
-        (0x1F1E6..=0x1F1FF).contains(&code) // Regional indicators
-    }) {
-        process_emoji_sequences(&normalized)
-    } else {
-        normalized
-    };
+    // An emoji is deleted by the character pass below like any other symbol,
+    // and the spaces around it stay, which is what GitHub does: its filter
+    // keeps `[\p{Word}\- ]` and drops everything else before turning spaces
+    // into hyphens.
 
     // Security Step 4: Filter dangerous Unicode characters
-    let sanitized = sanitize_unicode(&emoji_processed);
+    let sanitized = sanitize_unicode(&normalized);
 
     // Step 5: Convert to lowercase
     let mut text = sanitized.to_lowercase();
@@ -229,9 +217,6 @@ fn heading_to_fragment_internal(heading: &str) -> String {
     // First remove em-dashes and en-dashes entirely
     text = text.replace(['–', '—'], "");
 
-    // Step 8: Emojis were already replaced with hyphens in process_emoji_sequences
-    // No further processing needed for emoji markers
-
     // Step 9: Special symbol replacements
     // Handle ampersand based on position and surrounding spaces
     // GitHub's behavior:
@@ -272,13 +257,9 @@ fn heading_to_fragment_internal(heading: &str) -> String {
         if c.is_ascii_alphabetic() || c.is_ascii_digit() || c == '_' || c == '-' {
             // Preserve letters, numbers, underscores, and hyphens
             result.push(c);
-        } else if c == '§' {
-            // Preserve our marker character
-            result.push(c);
         } else if code == 0x20E3 {
             // Preserve combining keycap for keycap sequences
             // Note: FE0F should only be preserved as part of a keycap, not standalone
-            // The keycap preservation is handled in process_emoji_sequences
             result.push(c);
         } else if code == 0xFE0F {
             // Only preserve variation selector if it's preceded by a keycap base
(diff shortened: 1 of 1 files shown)


stasson:
Thank you for the quick merge !

rvben:
Thanks for the issue and fix! 🚀

PR #838, opened 2026-08-31
feat: add MD089 single blank line before heading rule

DavidPandleton:
## Summary

Adds a new opt-in rule **MD089 - Single blank line before heading** that enforces exactly one blank line before every heading (ATX and Setext), catching both **0 blank lines** (heading glued to previous content) and **2+ blank lines** (excessive spacing). The first heading in the document is exempt by default.

This complements MD022 (which only enforces a *minimum* of one blank around headings and only checks "at least") by also flagging *excessive* blank lines, and is fully auto-fixable.

## Rule behavior

- **0 blank lines** before a heading -> flagged ("Expected 1 blank line before heading, found 0")
- **2+ blank lines** before a heading -> flagged ("Expected 1 blank line before heading, found N")
- **Exactly 1 blank line** -> clean
- **First heading in document** -> exempt (configurable via `first-heading-exempt`)
- Headings inside fenced code blocks and front matter -> ignored
- Single-line HTML comments between content and heading -> treated as transparent

## Configuration

```toml
[global]
extend-enable = ["MD089"]

[MD089]
heading-styles = ["atx"]        # "atx", "setext", or both (default: both)
first-heading-exempt = true     # default: true
```

## Files changed

- `src/rules/md089_single_blank_line_before_heading.rs` - rule implementation (check + fix)
- `src/rules/md089_single_blank_line_before_heading/md089_config.rs` - serde config (kebab-case)
- `src/rules/mod.rs` - module registration + RULES entry (opt-in)
- `src/config/registry.rs` - RULE_ALIAS_MAP entries (`MD089` + `single-blank-line-before-heading`)
- `docs/md089.md` - full rule documentation
- `docs/rules.md` - Heading Rules table row
- Various docs - RULE_COUNT sentinels synced (82 -> 83)
- `tests/` - 22 unit tests, integration guards updated

## Testing

- 22 unit tests covering detection (0/1/2/3 blanks), first-heading exemption, style filtering, code-block safety, setext headings, and fix idempotency
- Full lib test suite: 7206 passed, 2 pre-existing perf-timing failures (text_reflow, unrelated)
- All CI docs/schema/code-block-tools gates pass
- `make check-rule-docs` clean: 83 rules (30 beyond markdownlint, max MD089)
- `cargo clippy` clean

Closes: #838

PR #865, opened 2026-09-09
feat(md092): new opt-in rule for inline formatting in headings

### LINKED ISSUE #864: New rule: forbid inline formatting in headings (opt-in)
### Problem or motivation

A heading is not only text on the page. It is also the source of generated artifacts: tables of contents, anchors and slugs, sidebar and outline entries. Inline markup inside a heading survives into those artifacts inconsistently — one generator strips the markers, another emits them literally, a third keeps the markup but drops it from the anchor — so the same heading can read `Method map()` in the page and `` Method `map()` `` in the TOC.

The second half of the problem is that markup in a heading is frequently unintentional. A heading naming a file or an identifier that contains `__` or `*` silently becomes emphasis:

```markdown
## __tests__/gt.test.js
```

renders as **tests**/gt.test.js — the underscores are gone from the heading, from the anchor and from the TOC, and the document is lint-clean, so nothing tells the author the path on the page is now wrong.

There is some in-repo evidence that markup inside headings already causes friction: #228 (MD063 does not consider inline code starting a heading) had to teach the capitalization rule about code spans in headings. A rule that lets a project forbid the markup outright removes that class of problem at the source.

### Proposed solution

A new opt-in rule — disabled by default and enabled with `extend-enable`, like MD088, MD089 and MD091 — that reports inline formatting inside a heading. Working name `no-formatting-in-headings`; the ID is yours to assign, I have MD092 in the draft to stay clear of the MD089 collision in #838.

Scope: `inlineCode`, `strong` and `emphasis` inside ATX and setext headings, at any nesting depth.

Three independent booleans, all `true` once the rule is enabled:

```toml
[global]
extend-enable = ["MD092"]

[MD092]
code = true      # `identifier` in a heading
strong = true    # **bold** in a heading
emphasis = true  # *italic* in a heading
```

They are separate because the three have different justifications, and a project that wants only bold banned should not have to fork the rule.

**No autofix, deliberately.** Unwrapping the markers changes what renders, rather than only how the source looks:

- `` ## `__tests__/gt.test.js` `` — dropping the backticks turns the path into bold text;
- `## **tests**/sort.test.js` — dropping the emphasis yields `tests/sort.test.js`, a plausible path that was never in the source.

Both are silent corruptions, so the rule reports and leaves the choice (rewrite the heading, escape the markers, move the identifier out of the heading) to the author.

### Alternatives considered

- **MD036** is the inverse case: emphasis used *instead of* a heading.
- **MD063** looks at headings but only at capitalization.
- **MD081** counts emphasis density across a document, without regard to whether the span sits in a heading.
- **Outside rumdl** this needs a second markdown parser in the pipeline for one check, and it then disagrees with rumdl's own parse of the same file.

As far as I can tell markdownlint has no equivalent rule, so this would be a rumdl-only rule, which is another reason for it to be off by default.

### Feature category

New lint rule

### Example markdown

```markdown
## Method `map()`

### **Practice**

## __tests__/gt.test.js

## The _.env_ file

Heading with `code`
-------------------
```

All five headings are reported when the rule is enabled with its defaults.

### Contribution

### Additional context

Happy to follow whatever ID and naming you prefer; renumbering the draft is a one-line change.

fey:
## Description

Adds MD092 (`no-formatting-in-headings`), an opt-in rule that reports an inline
code span, strong emphasis or emphasis inside a heading, in both ATX and setext
form. Closes #864.

A heading also feeds generated artifacts — a table of contents, an anchor, an
outline entry — and inline markup reaches those inconsistently: one generator
strips the markers, another emits them literally, a third keeps the markup in
the page and drops it from the anchor. The markup is also often unintentional:
`## __tests__/gt.test.js` is strong emphasis, so the underscores disappear from
the page, the anchor and the table of contents while the document lints clean.

Three decisions worth reviewing:

- **Off by default** (`extend-enable`), like MD088, MD089 and MD091. A code span
  naming a method in a heading is normal practice in many projects.
- **One switch per construct** (`code`, `strong`, `emphasis`, all `true` once the
  rule is enabled). The three shapes have different justifications, so a project
  that wants only one of them reported does not have to take the others.
- **No autofix, deliberately.** Removing the markers changes what renders rather
  than only how the source reads: dropping the backticks around a path in a code
  span turns the path into bold text, and dropping the emphasis from
  `## **tests**/sort.test.js` yields `tests/sort.test.js`, a path that was never
  in the source. Which remedy the author meant is not derivable from the source,
  so the rule reports and stops.

Markup inside a link's text in a heading is reported (the link text is what a
table of contents carries); the link itself is not. Both cases are pinned by
tests.

The rule ID is yours to assign — I took MD092 to stay clear of the MD089
collision in #838, and renumbering is a one-line change.

## Type of Change

- [ ] Bug fix (non-breaking change fixing an issue)
- [x] New feature (non-breaking change adding functionality)
- [ ] Documentation update

## Testing

- [x] Tests added/updated — 18 tests in
  `src/rules/md092_no_formatting_in_headings/tests.rs` plus two config tests
- [x] All tests passing

Four existing guards needed updating with the new rule, and they are in the
diff: the rule count and the frozen opt-in set in
`tests/integration/rules_mod_test.rs`, the configurable-rule count in the same
file, and the CLI/LSP coverage content map in
`tests/cli/cli_lsp_fix_consistency.rs`. Doc sentinels were regenerated with
`python3 scripts/check-rule-docs.py --write`.

On the full `cargo test` run, four tests fail on my machine and none of them
touch this change: two wall-clock assertions (`text_reflow` and list-block
performance tests that assert `< 100ms` in a debug build — a different pair
fails on each run) and two `code_block_tools` tests that need prettier
installed. Everything else passes. Built and tested with Rust 1.95 rather than
the pinned 1.96.

Validated against a large private corpus as well: 272 findings across 128 of
6115 files, with no false positive found by hand in a sample of them.

## Checklist

- [x] Code follows project style (`cargo fmt`, `cargo clippy --workspace --lib
  --tests` clean for the new code)
- [x] Documentation updated — `docs/md092.md`, plus the rule tables in
  `docs/rules.md` and the opt-in list in `docs/stability.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rvben:
Thanks @fey for contributing the heading-formatting rule and its thorough tests! I integrated it as MD093 because MD092 was assigned while this PR was open. The rule is now merged into main. I appreciate your work and patience with the renumbering.

PR #867, opened 2026-09-10
fix(MD057): name the missing index.md in roots-mode directory link failures

### LINKED ISSUE #863: MD057: trailing-slash absolute directory link reports a misleading "not found", and requires index.md outside a docs site
### rumdl version

0.2.69

### Installation method

Pre-built binary (via `uvx rumdl@0.2.69`, PyPI wheel)

### Bug description

With `absolute-links = "relative_to_roots"`, a repository-absolute link to an existing directory is accepted without a trailing slash (`/adir`) but rejected with one (`/adir/`).
The equivalent relative link (`../adir/`) is accepted either way.

I understand from [`resolve_under_root_with_opts`](https://github.com/rvben/rumdl/blob/749a7ea75b4bd73c4d8c2392d63c65eddefb490b/src/rules/md057_existing_relative_links.rs#L906) that this is deliberate: a trailing slash marks the destination as a directory-style link, which must serve `index.md`.
Two things still seem worth changing.

**1. The reported message does not describe the actual cause.**
Internally the per-root resolution returns `Resolution::DirectoryWithoutIndex`, but `relative_to_roots` collapses every failure across the roots into:

```
Absolute link '/adir/' was not found under any configured root or the project root
```

The docs-dir mode (`relative_to_docs`) has a precise message for this case — "resolves to directory '…' which has no index.md" — but the roots mode discards it.
The result sends you looking for a missing directory that is in fact present, which is what happened to me.
Even if the rule's verdict stays as it is, surfacing the `DirectoryWithoutIndex` reason when at least one root resolved to a directory would make it actionable.

**2. The `index.md` convention does not hold outside a static-site generator.**
The rationale in the code is MkDocs/Hugo URL routing, where `/section/` serves `section/index.md`.
In a plain repository rendered by GitHub, `/adir/` and `/adir` both resolve to the directory listing, and neither needs an `index.md`.
So under the default (non-MkDocs) flavor, the trailing slash carries no meaning that would justify the stricter check, and it produces a false positive on a link that works.

Related: #632 fixed the no-trailing-slash form of this in v0.1.96 ("MD057 now accepts absolute links that resolve to an existing directory"), which is what makes the remaining trailing-slash difference feel like an inconsistency rather than a deliberate distinction.

### Steps to reproduce

```
/tmp/repro/
├── .rumdl.toml
├── adir/some.md
└── docs/t.md
```

`.rumdl.toml`:

```toml
[MD057]
absolute-links = "relative_to_roots"
roots = ["."]
```

`docs/t.md`:

```markdown
# T

[abs dir no slash](/adir)

[abs dir slash](/adir/)

[rel dir slash](../adir/)
```

Run `rumdl check .` from `/tmp/repro`.

### Expected behavior

Under the default flavor, all three links resolve to the existing `adir/` directory and none is reported.

Failing that: if `/adir/` is to stay an error, the message should name the reason, as `relative_to_docs` already does.

### Actual behavior

```
docs/t.md:5:17: [MD057] Absolute link '/adir/' was not found under any configured root or the project root

Issues: Found 1 issues in 1/2 files (10ms)
```

Only the trailing-slash absolute form is flagged; `/adir` and `../adir/` both pass.

### Configuration

See the `.rumdl.toml` above.

### Operating system

Linux

### Additional context

Concretely, what I would expect is for the `index.md` requirement to follow the flavor rather than the trailing slash:

- `flavor = "mkdocs"` / `absolute-links = "relative_to_docs"`: keep the current behavior, since the site really does route `/section/` to `section/index.md`.
- Default flavor with `relative_to_roots`: treat a trailing slash as filesystem-equivalent, matching how relative links already resolve via `path.exists()`.

If you would rather keep the current semantics, an opt-out (something like `directory-links = "filesystem" | "index"`) would work just as well for our case.
We hit this while replacing lychee with `MD057` in a repository whose docs link to directories such as `/ansible/`; the workaround was to rewrite those links without the trailing slash.

Happy to open a PR for whichever shape you prefer.

jadhavgaurav:
## Description

With `absolute-links = "relative_to_roots"`, a link like `/adir/` that resolves
to a directory without an `index.md` was reported as:

```
Absolute link '/adir/' was not found under any configured root or the project root
```

which reads as though `/adir/` does not exist at all, even though the
directory is right there. The per-root resolution already distinguishes this
case internally (`Resolution::DirectoryWithoutIndex`), and the `relative_to_docs`
(MkDocs) variant of this check already reports it with a precise message
("resolves to directory '...' which has no index.md"). `relative_to_roots`
discarded that information and fell back to the generic message.

This PR keeps the verdict (the link still fails) and only fixes the message:
`validate_absolute_link_via_roots` now remembers the first
`DirectoryWithoutIndex` result across the configured roots and the
project-root fallback, and reports it with the same wording the docs_dir
variant already uses.

The second part of the issue (whether the `index.md` requirement should apply
outside MkDocs at all) is a separate design question and is left untouched
here.

## Type of Change

## Testing

- [x] Tests added/updated

Added two tests: one with an empty `roots` list (falls back to the project
root) and one with an explicit `roots = ["."]` entry, matching the
configuration in the original report. Verified both fail with the old message
by reverting only the source change and re-running; they pass again with the
fix restored.

## Checklist

- [x] Documentation updated (if needed): no user-facing docs describe the old
      wording, so none needed updating.

Fixes #863.

### DIFF
--- a/src/rules/md057_existing_relative_links.rs
+++ b/src/rules/md057_existing_relative_links.rs
@@ -839,24 +839,38 @@ impl MD057ExistingRelativeLinks {
     fn validate_absolute_link_via_roots(url: &str, roots: &[String], project_root: &Path) -> Option<String> {
         let (decoded, is_directory_link) = Self::prepare_absolute_url(url);
 
+        // A root that resolves the link to an existing directory lacking index.md
+        // is the actual cause when nothing else matches: remembered here so the
+        // fallback message can name it instead of claiming the target is missing.
+        let mut directory_without_index: Option<PathBuf> = None;
+
         for root in roots {
             let root_path = Self::resolve_against_project_root(root, project_root);
             // Filesystem mode: an existing directory without trailing slash is valid.
             // `require_index_for_dirs = false` aligns with relative-link behavior. (#632)
-            if matches!(
-                Self::resolve_under_root_with_opts(&root_path, &decoded, is_directory_link, false),
-                Resolution::Found
-            ) {
-                return None;
+            match Self::resolve_under_root_with_opts(&root_path, &decoded, is_directory_link, false) {
+                Resolution::Found => return None,
+                Resolution::DirectoryWithoutIndex { resolved } => {
+                    directory_without_index.get_or_insert(resolved);
+                }
+                Resolution::NotFound { .. } => {}
             }
         }
 
-        if matches!(
-            // Filesystem mode: see above.
-            Self::resolve_under_root_with_opts(project_root, &decoded, is_directory_link, false),
-            Resolution::Found
-        ) {
-            return None;
+        // Filesystem mode: see above.
+        match Self::resolve_under_root_with_opts(project_root, &decoded, is_directory_link, false) {
+            Resolution::Found => return None,
+            Resolution::DirectoryWithoutIndex { resolved } => {
+                directory_without_index.get_or_insert(resolved);
+            }
+            Resolution::NotFound { .. } => {}
+        }
+
+        if let Some(resolved) = directory_without_index {
+            return Some(format!(
+                "Absolute link '{url}' resolves to directory '{}' which has no index.md",
+                resolved.display()
+            ));
         }
 
         let msg = if roots.is_empty() {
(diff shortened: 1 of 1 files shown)


rvben:
Thanks for improving the diagnostic and adding focused tests. Since this PR was opened, roots/filesystem mode changed to accept an existing directory without `index.md`, including trailing-slash links. That means the proposed warning no longer applies there, so I’m closing the PR.

PR #870, opened 2026-09-11
feat(md034): add a config option to turn off auto fixes

guillp:
Add a config option in MD034 to turn off automatic fixes while keeding the rule enabled to report violations (the default is still enabled, as before).

This can be useful in at least 2 situations:
- when working with a mix of Markdown and MDX files: MDX compilation of embedded Mardown files breaks when rumdl autofixes .md by surrounding urls with `<` and `>`, which MDX doesn't support.
- You don't like long URLs inside your prose and usually want to hide them behind a human-readable text. This reports those URLs and let's you aply your own fix.

The diff is huge and touches many files since `MD034NoBareUrls` did not have a config before this change, so it now needs `::default()` everywhere it is instanciated.

### DIFF
--- a/benches/fix_performance.rs
+++ b/benches/fix_performance.rs
@@ -311,7 +311,7 @@ fn bench_fix_performance(c: &mut Criterion) {
 
     // MD034 - Bare URLs
     c.bench_function("MD034 fix", |b| {
-        let rule = MD034NoBareUrls;
+        let rule = MD034NoBareUrls::default();
         b.iter(|| rule.fix(black_box(&ctx)))
     });
 
--- a/docs/md034.md
+++ b/docs/md034.md
@@ -52,13 +52,21 @@ Chat with me at <xmpp:user@example.com>
 
 ## Configuration
 
-This rule has no configuration options.
+The `fix` option controls whether automatic fixes are emitted. It defaults to `true`.
+This can be useful when working with Markdown files that are embedded in MDX, since automatic fixes
+may break MDX compilation.
+
+```toml
+[MD034]
+fix = false
+```
 
 ## Automatic fixes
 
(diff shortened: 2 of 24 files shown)


rvben:
Thanks for the PR, @guillp, and for laying out both use cases so clearly.

Before we add a new option, I want to check whether existing settings already cover what you need.

**Mixed Markdown and MDX.** You can tell rumdl which `.md` files get compiled as MDX:

```toml
[per-file-flavor]
"docs/**/*.md" = "mdx"
```

Under the MDX flavor, MD034 fixes a bare URL as `[https://example.com](https://example.com)`, not `<https://example.com>`, so the result compiles. It has done this since v0.2.71.

**Reporting bare URLs but fixing them by hand.** The global `unfixable` setting works for any rule:

```toml
[global]
unfixable = ["MD034"]
```

`rumdl check` still reports every bare URL, and `rumdl fmt` leaves them alone.

Have you tried these? If they fall short for your setup, what's missing? That would tell me whether MD034 needs its own switch or whether something in these settings needs fixing.

One thing I found while looking into this: the editor's "Fix all" action ignored `unfixable`, even though `rumdl fmt` and format-on-save respected it. That's fixed for the next release, so if that's where you ran into trouble, it was a bug on our side.

guillp:
hey @rvben 

Thanks for the advices!

```
[global]
unfixable = ["MD034"]
```
does exactly what I want. I just wasn't aware this existed earlier ^^'

I just looked at other rules like MD028 or MD036 which have a `fix` toggle and figured I would implement the same kind of feature for MD034. I notice now that those have `fix = false` by default which is why this toggle exists for those specific rules.

I close this, since it is not needed.

Thanks!

PR #874, opened 2026-09-15
fix: empty run reason breaks machine-readable formats

guillp:
Fix for a bug that was introduced in 6cf7b89 / 0.2.73:

When a `rumdl` run ends up empty, `rumdl` prints a human-readable explanation before it prints the empty tests results. When ran with a machine-readable format (like `--output-format json`), this breaks the machine-readable output since both that notice and the machine readable empty output end up on stdout.

This fix prints the human-readable reason on stderr instead of stdout, only when a non-HR format is requested.

Before the fix:
```
% mkdir empty_dir
% rumdl check empty_dir 
No markdown files found to check.
% rumdl check empty_dir --output-format json 2>/dev/null
No markdown files found to check.
[]
```

After the fix:
```
% mkdir empty_dir
% rumdl check empty_dir 
No markdown files found to check.
% rumdl check empty_dir --output-format json 2>/dev/null
[]
```

Note that I have updated 2 tests with semantically equivalent code to make them easier to debug.


rvben:
Thanks for catching the broken JSON output and sending a fix. Current `main` keeps the empty-run notice separate from machine-readable output, so I’m closing this PR as superseded.

guillp:
Hey @rvben
Thanks for fixing this yourself. You might want to add the test changes from this PR since they are still relevant. Just let me know if you want me to send a PR, otherwise feel free to pick them up yourself from https://github.com/guillp/rumdl/tree/fix/broken_json_output_on_empty_run

rvben:
Hi @guillp, thanks for your work. I'll have a look at those test cases. 👍

PR #886, opened 2026-09-16
fix(MD057): find a link's destination when its text wraps onto another line

### LINKED ISSUE #878: MD057 misses relative links whose link text wraps onto a second line
### rumdl version

0.2.71

### Installation method

cargo install

### Bug description

MD057 (existing-relative-links) fails to detect a dead relative link when the link's text spans two physical lines inside a list item. Reproduced on both 0.2.47 and 0.2.71 with the default config.

Root cause (traced in the 0.2.71 source): in src/rules/md057_existing_relative_links.rs, both the contribute_dependency_targets pass (~line 384-403) and the main check pass (~line 1333-1362) do the same thing for each entry in ctx.links(): they index ctx.raw_lines()[link.line - 1], i.e. only the single line where the link's [ starts, then bail out early if that line doesn't contain the literal substring ](:

let line = lines[line_index];
if !line.contains("](") {
    continue;
}

When the link text wraps, the closing ]( lands on the next physical line, so this guard is always false and the link is skipped before any file-existence check runs. Since ctx.links() is built from a real parse, the link's destination is presumably already known accurately without needing to re-derive it via regex over a single raw line; gating on a single-line substring check seems avoidable.

Both call sites have this same guard, so it likely needs fixing in both places.

### Steps to reproduce

Repro (save as test.md, run rumdl check test.md):

  - Items reimbursable by the various [one-off
    expense](does-not-exist-anywhere)
    budgets.

  - Items reimbursable by the various [one-off expense](does-not-exist-anywhere) budgets.

### Expected behavior

Expected: MD057 flags does-not-exist-anywhere as missing.

### Actual behavior

With the test case, rumdl issues no warning at all. The same destination on a single line is caught correctly:

### Configuration

```toml
Default config.
```

### Operating system

Linux

### Additional context

_No response_

jmallach:
## Summary

- Both `check()` and `contribute_dependency_targets()` looked for a link's `](url)` only on `ctx.raw_lines()[link.line - 1]`, the single physical line where the link's `[` starts. When the link text wraps onto a later physical line, the closing `](` lands there instead, the `line.contains("](")` guard is always false, and the link is skipped before any existence check runs — so a dead destination goes unreported.
- `ParsedLink` already carries `end_line`, the physical line the parser says the link itself ends on, so the fix extends the scan window through it whenever it differs from the start line. `LINK_START_REGEX`'s `[^\]]` character class matches the embedded newline, so the existing bracket/URL-extraction regexes work unmodified over the extended window.
- Warning positions and fix ranges are now computed from the matched byte offsets via `LintContext::offset_to_line_col` instead of a hardcoded `link.line`, since a wrapped link's destination is not always on that line.

Fixes #878.

## Test plan

- [x] Added `test_missing_link_with_text_wrapped_across_lines` regression test
- [x] `cargo test --test lib` — 6046 passed, 0 failed
- [x] `cargo clippy --lib --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Manually verified: single-line links (no regression), multiple links on one line, a 3-physical-line wrap, and a wrapped self-referential link's fix all report/fix at the correct position


rvben:
Thanks for taking the time to trace this and put together a fix. The wrapped-link case is now handled on `main` in `943d1de9`, so I’m closing this PR as superseded.

PR #903, opened 2026-09-21
feat(encoding): lint files with non-UTF8 chars

guillp:
Allows rumdl to lint files with a low ratio of non-UTF8 characters, instead of rejecting them outright and not linting them at all.

The threshold for what constitutes a "low ratio" of non-UTF8 characters is configurable and defaults to 2.0%. Non-UTF8 characters beyond this threshold will still cause the file to be rejected, but with a linting finding instead of an error log.

Linting findings will be reported as usual, even for files with non-UTF8 characters within the allowed threshold. Findings will include issues related to non-UTF8 characters with their positions and context within the file.

Reports are not attached to dedicated rule but to a generic "encoding" diagnostic, since those findings happen before rules are even evaluated.

Implements #904


rvben:
Thanks @guillp for the PR, I'll try to review this today.

rvben:
Thanks @guillp, and thanks for checking in before going further. I'd like to get this into rumdl: today a non-UTF-8 file shows up as `[]` in JSON output, with only an error on stderr and exit code 2, which is easy to miss.

The config wiring is careful (provenance, `config get`, the parity and completeness tests). I'd still suggest a somewhat different shape, one that fits how rumdl handles its other pre-parse check:

- **A real rule, MD093**, modeled on MD092 (`src/merge_conflict.rs`). Register it as a rule and have its pre-parse check mirror `detect_configured` there, which applies `disable`, per-file-ignores, inline suppression and severity. Right now `--disable encoding` warns "Unknown rule" and nothing can silence the warning. It would also need a docs page like the other rules.
- **Always decode and lint.** Report each invalid sequence with its position, capped (say 20, then one summary warning). That covers the "too many" case, so the `non-utf8-threshold` key can go.
- **Decode with `fs::read` + `Utf8Chunks`.** It's simpler than the chunked reader, and it fixes a line-number drift I hit past the first 8 KB.
- **Cover `--stdin`** with the same decoder.
- **Exit code:** treat MD093 like any other finding (`check` follows `--fail-on`, `fmt` keeps its normal exit behavior), and unreadable files stay tool errors (exit 2). Replacing `invalid_utf8_file_is_a_tool_error` is fine under that model.

One thing is a hard requirement: **a file decoded lossily must never be written back.** On the current branch, if another fix applies to the same file, `--fix` / `fmt` saves the U+FFFD text, so a Latin-1 `é` is lost for good (and the encoding warning shows as `[fixed]`). A test that checks the bytes on disk are unchanged after `--fix`, on a file that also has a fixable issue, would pin this down.

That's a bigger change than you signed up for, so no pressure: if you'd rather stop here, I'm happy to build on your branch and keep your commits. Let me know which you prefer.

guillp:
Thanks @rvben , that is more or less the feedback I anticipated 😊
A dedicated rule makes sense, along with the rest of your comments.

Just one thing if I may: the threshold idea was to avoid linting a file that contains more binary than text, e.g. a binary file named .md by mistake. This kind of file is likely to trigger random rules by accident, potentially cluttering the ouput. Also, is there an upper cap to the parsed file size ? (I did not check).

Feel free to work on top or on the side of this PR, as you prefer. I'm off for the day and you most likely will produce something of better quality than me 👍

guillp:
Closing this, implemented by @rvben . Thanks again!

rvben:
Thanks again for raising the issue and the work on this PR that was central to bringing this to completion 💪😁

